PR #6, opened 2022-09-08
Copy tracing file in docker

s-laugh:
### Description

List of proposed changes:

This should fix the issue with the dast test; error message states "cannot locate 'tracing.js'".
Was able to reproduce locally by runing docker; this fixed that.

Can build/run docker


shewood (approved):
Looks like this is needed

PR #15, opened 2022-09-09
updated from next template

shewood:
### Description

List of proposed changes:

Updated Next Template in a few places to My Service Canada Account.
Note that this is not for the h1 - that will presumably come from the Heading from DS when we add it. 
-
-


DaceyTom2 (approved):
Looks good!

PR #17, opened 2022-09-09
Added codeowners file

DaceyTom2:
### Description
By default Nrds will be added to reviewers from the codeowners file


shewood (approved):
Looks good

PR #19, opened 2022-09-12
Removed logger example from home

DaceyTom2:
Removed logger example
Description
Removed logging example from home page


s-laugh:
I'd actually like to keep this until other "real" logging is put in place, to help troubleshooting with dynatrace reading the logs.

s-laugh (approved):

PR #22, opened 2022-09-14
Merging from main to dev

DaceyTom2:
Changes snuck through to main


s-laugh:
I don't think this is needed, no changes and the Network Graph looks like it's including all changes in both branches.

DaceyTom2:
main is a commit ahead of dev

s-laugh:
The only commit it's behind is a merge from "dev" (what used to be "development") to main. 
image
I'm guessing you created dev from development at it's latest commit, so essentially you just re-named the branch. So this is only adding a reciprocal commit back to itself. It won't cause any issues, but it's not needed.

DaceyTom2:
I understand that it is trivial. I'm just unsure why you would want main a commit ahead of dev unless hot fixing

s-laugh:
After every sprint, when we merge to main from dev, main will always be a commit ahead, the merge commit. That's just how the branching works the way we've set it up. The only time we should be merging from main to dev is when there is a hot fix.

PR #23, opened 2022-09-15
update action triggers

s-laugh:
### Description

List of proposed changes:

- Update workflow triggers to match [CI/CD plan](./../wiki)
- Update load trigger to `workflow_dispatch` (as it's always failing, but can still be manually triggered)


s-laugh:
Expecting to close this in place of #27 if approved

PR #26, opened 2022-09-15
Tests only Node v18.x

s-laugh:
### Description

List of proposed changes:

- Update workflow matrix to only test against node v18.x

Verify no tests were run against v16.x in this PR

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16.x, 18.x]
+        node-version: [18.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
     steps:
       - name: Checkout 🔔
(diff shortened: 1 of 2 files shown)


DaceyTom2 (approved):
This is a much better solution than #27

s-laugh:
> This is a much better solution than #27

Unfortunately it's really only a band-aid, not a solution.

DaceyTom2:
I'm not sure how it is a band-aid. SCH will be deployed using the latest version of node (currently 18.9). We should be aiming to test as close to production as possible

DaceyTom2:
The conflict only arises because the same test is committing to the same file in the same branch. There is no situation where we want this to happen as it overwrites the results of another test

s-laugh:
That's actually not true, read the logs https://github.com/DTS-STN/secure-client-hub/actions/runs/3068273207/jobs/4955539103#step:9:74
It's failing while trying to force push because from the time the script checks out the branch to the time it goes to push the changes, there is a new commit on the branch. This same issue will happen from time to time, no matter what we do; however by spiting the deploy to gh-pages and the tests we can easily re-run the deploy job separately rather than re-running the whole test. It also lets us merge without requiring the deploy, because the test will still show as passed. In addition, since e2e and unit tests are always getting run at the same time, if we put those deploys together, we greatly reduce the chances of a commit conflict. That's exactly what #27 is doing.

PR #27, opened 2022-09-15
72252 deploy pages 1commit

s-laugh:
### Description

List of proposed changes:

- Moves default tests to a single workflow
  Enables us to store and download artefacts in a later job
  This lets us push to gh-pages with a single commit (less error prone)
  Less likely to fail a test because the deploy failed
- Removes the use of matrix (we aren't using it)
- updates the triggers for other workflows

This will complete work purposed in #26 #23 

- Tests still execute as expected according to our CI/CD plan.
- A single commit with both e2e & unit results is added to gh-pages

### Additional Notes

This has a better visual representation of the tests, seen here https://github.com/DTS-STN/secure-client-hub/actions/runs/3062704688


.github/workflows/dast.yml
```diff
@@ -1,8 +1,10 @@
 name: Dynamic Application Security Tests
 
-on:
+on:  
+  pull_request:
+    branches: [release*]
   push:
-    branches: [main]
+    branches: [main, release*]
```
DaceyTom2, line 7: It would be better to pick either "on PR" or "on push". Running the workflow multiple times is redundant

.github/workflows/default-tests.yml
```diff
@@ -0,0 +1,146 @@
+name: Default Testing
+
+on: 
+  pull_request:
+    branches: [dev, main, release*]
+  push:
+    branches: [dev, main, release*]
```
DaceyTom2, line 7: Same as above

.github/workflows/default-tests.yml
```diff
@@ -0,0 +1,146 @@
...
+      with:
+        path: test_results
+      env:
+        CI: true
+
+    - name: Deploy Report 🚀
+      uses: JamesIves/github-pages-deploy-action@v4
+      with:
+        target-folder: ${{ github.ref }}
+        folder: test_results
```
DaceyTom2, line 146: These tests are easier to maintain in separate files. Publishing to gh-pages was failing because multiple of the same test were running concurrently publishing the same file. This was caused by using node v16 and v18 splitting the tests

DaceyTom2 (changes requested):

.github/workflows/dast.yml
s-laugh: This rational is a bit complex, but I'm open to changing this... Let me explain

We want our DAST to run **before** we push to production, so on push to main makes perfect sense for the happy path. The complexity is when it comes to a hotfix, if the test only runs after code get to the release branch it could be released to production without the test passing, so that's why I added the on pull_request. However running on pull_request, doesn't test post merge, meaning their could be discrepancies that cause issues, so then we'd want to run it again post merge into a release branch, hence on push to release*.

I'm not sure of a better way to handle this, we might want to talk about it more with the diagram in front of us https://github.com/DTS-STN/secure-client-hub/wiki.

.github/workflows/default-tests.yml
s-laugh: This is actually required for codeQL to work, it need the test to be run on the branch (push) that it's merging into so that it has something to compare against when it's run on pull_request, where we need it for proper "gating" of our branches.

Here also, testing prior to merge, doesn't always mean things are guaranteed to work post merge, so having it in both helps shorten feedback loops.

.github/workflows/default-tests.yml
s-laugh: I agree that maintaining in separate files can be easier; however doing it this way is incredibly powerful and gives us a lot more control over the testing, reducing some complexity. It's also not unmaintainable, only ~150 lines not 1000s.

Publishing to gh-pages failed more often because of a conflict in the branch; this happened because multiple commits were being pushed at the same time. This happened in our sc-digital-centre repo regularly where there was only one test for each. Removing the matrix will help a bit, while this helps even more.

.github/workflows/dast.yml
DaceyTom2: I don't remember if Zap fails a PR if issues are found (perhaps just new issues). The process is a bit more manual in which it opens an [issue](https://github.com/DTS-STN/secure-client-hub/issues/10) with the project and describes what it finds for devs to review.

In the scenario of a hotfix or regular release we should be checking what Zap reports before clicking publish within TC. Another factor to consider is, if prod is down, we likely wont want to be waiting on ZAP to do a 5 minute scan while hotfixing.

s-laugh: I absolutely agree with your last point; I'll remove the pull_request.

DaceyTom2 (approved):
Github is glitching and I can't see responses to the ongoing conversation.
We agreed to combine the workflows and see how it performs

PR #29, opened 2022-09-16
72246 implement graphql client

DaceyTom2:
### Description
GraphQL client boilerplate code added to SCH

List of proposed changes:
- Added AEM env variable
- Added GraphQL client
- Updated dockerfile to add env

Currently unimplemented in the code

### Additional Notes
Teamcity will need to be updated to include the env. It currently exists in vault


Dockerfile
```diff
@@ -9,10 +9,8 @@ FROM base AS build
 # Build envs
 ARG BUILD_DATE
 ENV BUILD_DATE=$BUILD_DATE
-ARG ENV_EXAMPLE
-ENV ENV_EXAMPLE=$ENV_EXAMPLE
-ARG NEXT_PUBLIC_ENV_EXAMPLE
-ENV NEXT_PUBLIC_ENV_EXAMPLE=$NEXT_PUBLIC_ENV_EXAMPLE
+ARG AEM_GRAPHQL_ENDPOINT
+ENV AEM_GRAPHQL_ENDPOINT=$AEM_GRAPHQL_ENDPOINT
```
s-laugh, line 13: If the env variable is added at runtime, pretty sure we determined it's not ALSO required a build...

graphql/client.js
```diff
@@ -0,0 +1,23 @@
...
+  cache: new InMemoryCache(),
+})
+
+export default async function (AEMQuery) {
+  return client
+    .query({
+      query: AEMQuery,
+    })
+    .catch((error) => {
+      console.log(error)
```
s-laugh, line 21: should use our logger here

s-laugh (changes requested):

nibivi77 (approved):

nibivi77 (commented):
Approved before seeing Shaun's comments. Good points were raised with both comments.

Dockerfile
s-laugh: ahh good point, so this env var might be an exception to that "rule"

graphql/client.js
DaceyTom2: This catch needs to be able to fail hard once we get a working example. I'll leave it for now

Dockerfile
DaceyTom2: It's not a rule parse, different environment variables have different purposes. As long as the variable isn't prefixed with next_public they should be relatively safe

graphql/client.js
s-laugh: Failing hard is fine; we should still log it properly.

DaceyTom2: It is up to the parent caller to log so the full stack trace can be seen

s-laugh: If I'm understanding what you're suggesting correctly, you want each call to AEM to handle it's errors individually so that more of the stack trace can be seen. Assuming the goal would be to know which call failed, why not just include the AEMQuery in the log, or pass an ID and put that in the log, which would reduce the duplication of having to log in multiple locations and also risk having each done differently.

s-laugh (approved):
Agreed, catching, fail hard flow will resolve once implemented.

PR #34, opened 2022-09-20
Sch 73761

shewood:
### Description
Update the tailwind config file

List of proposed changes:
Updated the tailwind config based on previous settings for the similar project. 
updated a few colours but left most as we don't have all of the ones that we will be using yet and we don't want to add the entire lot of DS colours at this point. 
-


DaceyTom2 (approved):
LGTM

PR #38, opened 2022-09-26
Adding tag with Cypress-Grep.

nibivi77:
### Description

Added a tag to one of our cypress tests to set up Smoke testing.
Added a step to our default-test.yml file to run the Smoke tests before the rest of the e2e tests.

Documentation can be found here:

[Multiple methods, Method 3 is Cypress-Grep](https://testersdock.com/cypress-test-tags/)

[Cypress-grep npm](https://www.npmjs.com/package/cypress-grep)

### To run
split the terminal, the first terminal, `npm run dev`, and in the second terminal, `npx cypress run --env grepTags="smoke"`

### Result
Three will be pending because they do not have the `smoke` tag.

image

### DIFF
--- a/.github/workflows/default-tests.yml
+++ b/.github/workflows/default-tests.yml
@@ -77,6 +77,15 @@ jobs:
         env:
           CI: true
 
+      - name: Smoke Tests 💨
+        uses: cypress-io/github-action@v4
+        with:
+          install: false
+          env: grepTags=smoke
+          record: false
+          tag: smoke
+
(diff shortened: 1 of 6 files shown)


shewood (changes requested):
It looks like there is an issue with the actions where none are being called. 
Status
Startup failure

nibivi77:
We have decided to use regular Cypress tests as our Smoke tests considering Dynatrace will play a big part in our pipelines.

PR #39, opened 2022-09-27
Renamed sch to secure-client-hub - reverting pr 28

DaceyTom2:
### Description
Reverting sch back to secure-client-hub

List of proposed changes:
- helm files and directories renamed to secure-client-hub

Ensure app deploys as secure-client-hub.[env]...


P0NDER0SA (approved):
yep, this looks right

PR #42, opened 2022-10-03
Add fonts 78640

shewood:
### Description
added font awesome 
List of proposed changes:
added font awesome so that the fonts can be dynamically added. 
added security header change so that the fonts and icons appear from Design system.


DaceyTom2 (approved):
Looks good!

PR #50, opened 2022-10-04
Card Component

nibivi77:
### Description

Added Card Component.

### Additional Notes
- Tests are coming in a separate PR.

### DIFF
--- a/__tests__/pages/home.test.js
+++ b/__tests__/pages/home.test.js
@@ -5,7 +5,6 @@ import { render, screen } from '@testing-library/react'
 import '@testing-library/jest-dom'
 import Home from '../../pages/home'
 import { getStaticProps } from '../../pages/home'
-import en from '../../locales/en'
 
 import { useRouter } from 'next/router'
 
@@ -36,8 +35,8 @@ describe('Home page', () => {
 
   it('should render the page', () => {
     render(<Home locale="en" content={content} />)
-    const heading = screen.getByText(en.pageHeading.title)
-    expect(heading).toBeInTheDocument()
+    const homeDiv = screen.getByTestId('homeContent-test')
+    expect(homeDiv).toBeInTheDocument()
   })
 
   it('Test getStaticProps', async () => {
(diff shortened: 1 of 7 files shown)


__tests__/pages/home.test.js
```diff
@@ -35,7 +36,7 @@ describe('Home page', () => {
 
   it('should render the page', () => {
     render(<Home locale="en" content={content} />)
-    const heading = screen.getByRole('heading')
+    const heading = screen.getByText(en.pageHeading.title)
```
s-laugh, line 39: This is going to have to change again when we move to the CMS; is there a more permanent fix maybe? (We don't care that the heading exists per say, but that the page rendered...)

pages/home.js
```diff
@@ -13,6 +14,7 @@ export default function Home(props) {
     <div id="homeContent">
       <Heading id="my-dashboard-heading" title={t.pageHeading.title} />
       <p className="py-8">{props.content.paragraph}</p>
+      <Card t={t} />
```
s-laugh, line 17: I'm conflicted over the simplicity of just passing the single prop for all text vs. specifying each one... not sure if there are other thoughts; if this is a pattern we want to continue?

tailwind.config.js
```diff
@@ -45,6 +45,7 @@ module.exports = {
           light: '#5E8EBD',
           medium: '#173451',
           dark: '#21303F',
+          button: '#295376',
```
s-laugh, line 48: Should we be adding colours outside the defined pallet, or seek correction from the design team members?

s-laugh (commented):
Just posing some questions (food for thought); not blocking. 
Otherwise it looks great!

__tests__/pages/home.test.js
nibivi77: Yea I was thinking about that. I would of used data-test-id by Design System does not support it. I have an idea.

pages/home.js
nibivi77: I am fine with both, I'll change it up if you prefer.

tailwind.config.js
nibivi77: I don't know where the pre-defined color pallet comes from. I don't mind adding colors to it. I will raise this to the stand-up tomorrow.

s-laugh (approved):
Great work!

PR #51, opened 2022-10-04
New Test scripts to validate dashboard page title_73199

manishasharma5757:
### Description

- Added test case to validate the Dashboard page heading in English
- Added test case to validate the Dashboard page heading in French

Dashboard Page heading in French and English


nibivi77:
I see nothing wrong here. Just a question, why do we use two files to return simple `cy.get()` objects? Just curious. :)

nibivi77 (approved):
Fair enough! Looks good :)

PR #52, opened 2022-10-05
Card and ViewMoreLess Unit Tests

nibivi77:
Added simple unit tests to cover `Card.js` and `ViewMoreLessButton.js`.


DaceyTom2 (approved):
Tests look good!

PR #53, opened 2022-10-05
Commiting smoke tests.

nibivi77:
Draft for Smoke tests and as an example.


nibivi77:
cypress dashboard needs membership in order to use it.

PR #55, opened 2022-10-05
Test cases added for 72243

manishasharma5757:
### Description
- Added test case to validate that the Card placeholder is present on dashboard page
- Added test case to validate that the Card placeholder has a title/heading
- Added  test case to validate that the Test Card expands on clicking

Test Card placeholder is present on dashboard page


nibivi77:
I see nothing wrong here. Just a question, why do we use two files to return simple cy.get() objects? Just curious. :)

manishasharma5757:
> I see nothing wrong here. Just a question, why do we use two files to return simple cy.get() objects? Just curious. :)

I am using Page objects to make test cases more readable and clean.

PR #59, opened 2022-10-11
Task list

shewood:
### Description
Added the task list

List of proposed changes:
For now, added the task list as a file (later this will presumably come from AEM) 
Added the benefit tasks and the most requested tasks component so that each part can be styled separately. 
Ben added the logic to display the programs/tasks and make the most requested task separate from the other tasks 

Note that the styling is not finalized yet and also the tasks listed are not exactly as shown in the figma as this content will be coming from AEM so the current work will be removed at that point.

### DIFF
--- a/components/Card.js
+++ b/components/Card.js
@@ -1,10 +1,19 @@
 import propTypes from 'prop-types'
 import ViewMoreLessButton from '../components/ViewMoreLessButton'
 import BenefitTasks from './BenefitTasks'
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 
 export default function Card(props) {
   const [isOpen, setIsOpen] = useState(false)
+  const [mostReq, setMostReq] = useState({})
+  const [tasks, setTasks] = useState([])
+
+  useEffect(() => {
+    const test = props.taskGroups.shift()
+    setMostReq(test)
+    setTasks(props.taskGroups)
+  }, [props.taskGroups])
+
   return (
     <div className="border rounded border-gray-300 shadow">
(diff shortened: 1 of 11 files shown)


components/Card.js
```diff
@@ -1,12 +1,17 @@
 import propTypes from 'prop-types'
 import ViewMoreLessButton from '../components/ViewMoreLessButton'
-import { useState } from 'react'
+import BenefitTasks from './BenefitTasks'
+import { useState, useEffect } from 'react'
```
nibivi77, line 4: We can remove `useEffect` since we are not using it :) (Sorry I forgot to remove it)

pages/home.js
```diff
@@ -4,19 +4,46 @@ import en from '../locales/en'
...
-      <p className="py-8">{props.content.paragraph}</p>
+
+      <Card
+        programUniqueId={'ei'}
+        locale={props.locale}
+        cardTitle={ei.programTitle}
+        viewMoreLessCaption={t.viewMoreLessButtonCaption}
+        taskHeading={ei.taskHeadingKey}
+        taskGroups={ei.tasksGroups}
+        // callout={MapCallout(value.statusCode, value.typeCode, t)}
```
nibivi77, line 28: We can also remove all the callout commented lines. :+1:

DaceyTom2:
BenefitTasks and MostReqTasks need test files to go along with the new components

components/Card.js
shewood: done

shewood:
I will add the tests in a new PR with the combined component.

nibivi77 (approved):
This looks good ! :+1:

PR #61, opened 2022-10-11
75022_E2E_TCs

manishasharma5757:
### Description

List of proposed changes:
TC to validate that the Task List is present 

validate that the Task List is present


nibivi77 (approved):
I like it! Thanks !

PR #65, opened 2022-10-13
added link for pages

shewood:
### Description

List of proposed changes:
Added text and link for navigating security and profile page
Added button to return to dashboard '/home' page
- the components are ready to be added in the commented out text that can then be removed once added to each page once completed.

Test will come later

### Additional Notes
Please check the router bit. 
Text will eventually come from AEM.

### DIFF
--- a/components/PageLink.js
+++ b/components/PageLink.js
@@ -1,0 +1,72 @@
+import PropTypes from 'prop-types'
+import Link from 'next/link'
+import { useRouter } from 'next/router'
+import { Button } from '@dts-stn/service-canada-design-system'
+
+export default function PageLink(props) {
+  const router = useRouter()
+  return (
+    <>
+      <div className="pt-2 md:pt-4 my-8 border-t border-gray-light ">
+        <h2 className="font-display font-bold text-[32px] md:text-[36px]">
+          {props.lookingForText}
+        </h2>
+        <div className="font-body mb-10  text-xl">
+          <span className=" text-gray-darker" id={`link-for-${props.linkText}`}>
+            {props.accessText}{' '}
+          </span>
+          <Link href={props.href}>
+            <a className="text-blue-default hover:text-blue-hover visited:text-purple-medium underline">
+              {props.linkText}
+            </a>
+          </Link>
+        </div>
+
+        <Button
+          id={props.buttonId}
+          onClick={() => router.push('/home')}
+          styling={props.styling}
+          text={props.buttonText}
+        />
+      </div>
+    </>
+  )
+}
+
+PageLink.propTypes = {
+  // Props for the text and link
+  lookingForText: PropTypes.string,
+  linkText: PropTypes.string,
+  accessText: PropTypes.string,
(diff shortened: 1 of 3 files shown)


components/PageLink.js
```diff
@@ -0,0 +1,72 @@
...
+import Link from 'next/link'
+import { useRouter } from 'next/router'
+import { Button } from '@dts-stn/service-canada-design-system'
+
+export default function PageLink(props) {
+  const router = useRouter()
+  return (
+    <>
+      <div className="pt-2 md:pt-4 my-8 border-t border-gray-light ">
+        <h2 className="font-display font-bold text-[32px] md:text-[36px]">
```
s-laugh, line 11: Not Blocking - I'm not sure its a good idea to use "Arbitrary values" if at all possible - we should be striving for consistency and this won't help us do that. Assuming this comes direct from the figma designs, I would suggest reviewing the font size with our designers to ensure we build with consistency, and maybe extend the font sizes in the tailwind config like we do for colors.

components/PageLink.js
```diff
@@ -0,0 +1,72 @@
...
+          {props.lookingForText}
+        </h2>
+        <div className="font-body mb-10  text-xl">
+          <span className=" text-gray-darker" id={`link-for-${props.linkText}`}>
```
s-laugh, line 15: the `linkText` could contain spaces, that shouldn't be included in an `id` property

components/PageLink.js
```diff
@@ -0,0 +1,72 @@
...
+          <Link href={props.href}>
+            <a className="text-blue-default hover:text-blue-hover visited:text-purple-medium underline">
+              {props.linkText}
+            </a>
+          </Link>
+        </div>
+
+        <Button
+          id={props.buttonId}
+          onClick={() => router.push('/home')}
```
s-laugh, line 27: Both links on this component (button and link) would be to internal pages, and we want to ensure we use proper next Routing for that... I believe using a button with an onClick like this isn't an accessible way of routing. I believe it should be a link, styled like a button.

components/PageLink.js
```diff
@@ -0,0 +1,72 @@
...
+PageLink.propTypes = {
+  // Props for the text and link
+  lookingForText: PropTypes.string,
+  linkText: PropTypes.string,
+  accessText: PropTypes.string,
+  href: PropTypes.string,
+  // Props for the Button
+  id: PropTypes.string,
+  buttonId: PropTypes.string,
+  styling: PropTypes.string,
```
s-laugh, line 45: Maybe want to give this prop a more specific name at the component level. Styling for what? Is it tailwind CSS or style type?

s-laugh (commented):
Just some thoughts for you :)

components/PageLink.js
shewood: The styling is coming from the DS. that is what they called if for the style of the button.

shewood:
Closing this in favour of a new PR.

PR #67, opened 2022-10-14
81763 home aem query

DaceyTom2:
### Description
Added query and mapper for AEM home

List of proposed changes:
- Query added
- Mapper added

From home the content object returned from mapHome can be logged

### Additional Notes
I haven't added test for the mapper yet as it is expected to change shortly
Mapper can also use some cleanup/refactoring and am looking for input from other developers

### DIFF
--- a/graphql/client.js
+++ b/graphql/client.js
@@ -1,4 +1,4 @@
-import { ApolloClient, InMemoryCache, HttpLink } from '@apollo/client'
+import { gql, ApolloClient, InMemoryCache, HttpLink } from '@apollo/client'
 
 const link = new HttpLink({
   uri: process.env.AEM_GRAPHQL_ENDPOINT,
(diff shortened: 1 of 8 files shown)


graphql/client.js
```diff
@@ -13,11 +13,9 @@ const client = new ApolloClient({
...
 
 export default async function (AEMQuery) {
-  return client
-    .query({
-      query: AEMQuery,
-    })
-    .catch((error) => {
-      console.log(error)
-    })
+  return client.query({
```
nibivi77, line 16: Should we have a `catch()` here? Good opportunity for our logger as well.

pages/home.js
```diff
@@ -47,7 +47,7 @@ export default function Home(props) {
 }
 
 export async function getStaticProps({ locale }) {
-  const content = await fetchContent()
+  const content = await getHomeContent()
```
nibivi77, line 50: Should we pass the content into the cards? or do this in a separate PR?

graphql/client.js
DaceyTom2: I'd rather not catch the error here since we won't know what page is throwing it. I'll add at the page level

PR #71, opened 2022-10-17
81770 Added profile card component

CurtisUnderwood:
### Description

List of proposed changes:

- Updated card component to add a flag for most-requested tasks
- Added cards with most-requested tasks off

### Additional Notes
- Tasks to be added in next PR

### DIFF
--- a/components/Card.js
+++ b/components/Card.js
@@ -6,8 +6,12 @@ import MostReqTasks from './MostReqTasks'
 
 export default function Card(props) {
   const [isOpen, setIsOpen] = useState(false)
+
+  var tasks = props.taskGroups
   const mostReq = props.taskGroups[0]
-  const tasks = props.taskGroups.slice(1, props.taskGroups.length)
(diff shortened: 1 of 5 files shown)


components/Card.js
```diff
@@ -6,8 +6,12 @@ import MostReqTasks from './MostReqTasks'
 
 export default function Card(props) {
   const [isOpen, setIsOpen] = useState(false)
+
+  var tasks = props.taskGroups
   const mostReq = props.taskGroups[0]
-  const tasks = props.taskGroups.slice(1, props.taskGroups.length)
+  if (props.mostReq) {
+    tasks = props.taskGroups.slice(1, props.taskGroups.length)
```
DaceyTom2, line 13: This is fairly confusing. I proposed an idea to @shewood that I think would work well

Remove this code and then below we can do
<MostReqTasks taskList={props.taskGroups} mostReq={true} />

PR #74, opened 2022-10-17
Bug fix Exit beta lowercase

DaceyTom2:


shewood (approved):
I tried it out and it works.

PR #82, opened 2022-10-19
73202_E2E_TCs

manishasharma5757:
### Description

List of proposed changes:
-Test case to Validate that the Task List is Present for each card on profile page
-Test case to Validate that each card has a "Most Requested" section
-Test case to Validate that each card has a "application" section

Testing Profile page task list and components using Cypress


cypress/e2e/SecuritySettings.cy.js
```diff
@@ -84,4 +81,53 @@ describe('Validate Security Settings page', () => {
...
+    dashboardPo.FrenchButton().click()
+    profilePo
+      .LookingFor()
+      .should('have.text', 'Vous recherchez les paramètres de votre profil?')
+    profilePo
+      .BackToDashboardButton()
+      .should('have.text', 'Retour à mon tableau de bord')
+  })
+
+  it.skip('Validate the "Vous recherchez les paramètres de sécurité?" click navigates to /fr/security Page', () => {
```
nibivi77, line 120: Any specific reason why we skip this one?

nibivi77 (commented):
The tests are good! The selectors are a little tricky because some of the selectors are based on css elements and html tags. That's not recommended for obvious reasons. For now, its good to me since the app will most likely change, and we will have to correct the cypress tests if the changes do happen.

cypress/e2e/SecuritySettings.cy.js
manishasharma5757: > Any specific reason why we skip this one?

This test case was failing, so Sheila added a new test-id to one of links. In next PR, I have removed the skip

PR #83, opened 2022-10-20
75736_E2E_TCs

manishasharma5757:
### Description

List of proposed changes:
- Added test cases to validate content on Dashboard page

validate content on Dashboard page using Cypress


cypress/e2e/PageObjects/ProfilePO.cy.js
```diff
@@ -1 +1,98 @@
...
+function Eachsectionheading() {
+  return cy.get('.pb-12 >div>div')
+}
+
+function MostReqheading() {
+  return cy.get('.pb-12 >div')
+}
+
+function Cards() {
+  return cy.get('#homeContent>div')
```
nibivi77, line 28: This is an example but instead of getting all divs inside the id `homeContent`, I would add a data-testId to the cards.

https://testing-library.com/docs/queries/bytestid/

nibivi77 (commented):
If you can, try to create data-testId's for selectors instead of getting items by css or by html elements.

cypress/e2e/PageObjects/ProfilePO.cy.js
nibivi77: If we add a random div that is not a card inside the `#homeContent`, then this function will return more than just cards.

manishasharma5757: I wanted to apply loop by finding multiple divs

manishasharma5757: > If we add a random div that is not a card inside the `#homeContent`, then this function will return more than just cards.

I wanted to apply loop by finding multiple divs

manishasharma5757: > This is an example but instead of getting all divs inside the id `homeContent`, I would add a data-testId to the cards.
> 
> https://testing-library.com/docs/queries/bytestid/

I wanted to apply loop by finding multiple divs

manishasharma5757:
> If you can, try to create data-testId's for selectors instead of getting items by css or by html elements.

yes,it's preferable not to use the css as it often changes. I will be doing it from now onwards.

DaceyTom2 (approved):
I'll approve but I agree with @nibivi77. If you add a test id to the parent you can loop over a list of children. It makes test cases more readable and less prone to errors

PR #84, opened 2022-10-20
Security query and mapper.

nibivi77:
### Description

- Added security query.
- Added security aem mapper.

### Notes

- Breadcrumb content is not being used from the query for now.
- The logic for the mapper is not the best and there are no checks for errors. Thats why we throw errors from the `security.js` page. Any thoughts on making it better?

### DIFF
--- a/graphql/mappers/security.js
+++ b/graphql/mappers/security.js
@@ -1,0 +1,104 @@
+import clientQuery from '../client'
+
+export async function getSecurityContent() {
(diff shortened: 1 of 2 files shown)


DaceyTom2 (approved):
Looking good,
In terms of error handing I'm not sure what the best approach is.

Some content missing or broken is likely okay but it is hard to say at what point we just redirect to a 500 error

PR #86, opened 2022-10-20
81779 Mapped profile content

CurtisUnderwood:
### Description

List of proposed changes:

- Added mapper for profile

### Additional Notes
- Existing issue with the task list component tasks all appear in one single column this will need a fix


lib/loadIcons.js
```diff
@@ -30,4 +30,13 @@ export const icon = {
...
   'user-check': solid('user-check'),
+  'money-bill-transfer': solid('money-bill-transfer'),
+  'phone-flip': solid('phone-flip'),
+  'location-dot': solid('location-dot'),
+  'circle-minus': solid('circle-minus'),
+  'circle-plus': solid('circle-plus'),
+  'address-book': solid('address-book'),
+  'map-location-dot': solid('map-location-dot'),
+  'comment-dots': solid('comment-dots'),
+  'bell-solid 1': solid('bell'),
```
DaceyTom2, line 41: It is worth putting these in alphabetical order so we can quickly see if icons are already added

pages/profile.js
```diff
@@ -4,25 +4,26 @@ import PageLink from '../components/PageLink'
...
-export default function Home(props) {
+export default function Profile(props) {
   /* istanbul ignore next */
   const t = props.locale === 'en' ? en : fr
 
   return (
     <div id="homeContent" data-testid="homeContent-test">
       <Heading id="my-dashboard-heading" title={t.pageHeading.profile} />
       {props.content.cards.map((card) => {
+        console.log(card.lists)
```
DaceyTom2, line 18: This looks like an oopsie

DaceyTom2 (approved):
Looks really good!

pages/profile.js
DaceyTom2: removed log

lib/loadIcons.js
DaceyTom2: ordered icons

PR #94, opened 2022-10-25
80962_TCs

manishasharma5757:
### Description

List of proposed changes:

- Created new test cases for Security Page to validate content and Links present on security page

New test cases to Validate security page using cypress


components/Card.js
```diff
@@ -14,7 +14,7 @@ export default function Card(props) {
   }
 
   return (
-    <div className="border rounded border-gray-300 shadow my-6">
+    <div className="border rounded border-gray-300 shadow my-6" data-cy="Cards">
```
shewood, line 17: Would it be ok to keep to lower case?

components/Card.js
```diff
@@ -35,14 +35,21 @@ export default function Card(props) {
...
-            <div className="bg-deep-blue-60d mt-4 pl-2">
+            <div
+              className="bg-deep-blue-60d mt-4 pl-2"
+              data-cy="most-requested-section"
+            >
               <MostReqTasks taskListMR={mostReq} dataCy="most-requested" />
             </div>
           )}
-          <div className=" md:columns-2 gap-8 pt-8">
+          <div className=" md:columns-2 gap-8 pt-8" data-cy="Task-List">
```
shewood, line 45: same here?

manishasharma5757: ok let me change.Thanks

nibivi77 (approved):
Same as the other PR. Great work. We are almost using the best practice recommended by cypress. 

This looks good for me Manisha! :)

components/Card.js
manishasharma5757: > same here?

ok,I will change it these to lower case.Thanks

components/Card.js
manishasharma5757: @shewood done!

cypress/e2e/dashboard.cy.js
```diff
@@ -68,243 +62,248 @@ describe('Validate dashboard page', () => {
...
       'Assurance-emploi',
       'Rapports et documents',
       '7'
     )
   })
 
-  it.skip('Validate the EI card "Reports and documents" section', () => {
-    profilePo.ValidateCardTaskListAndSection(
+  it('Validate the EI card "Personal information" section', () => {
+    dashboardPo.ValidateCardTaskListAndSection(
```
shewood, line 122: Would it be possible to combine these tests together so they check the sections all at once so that it runs faster?

manishasharma5757: Yes, going to create a create a PR soon for code optimization

PR #100, opened 2022-10-28
Bug: 82988 replace sctitle with sclinktext and sclinttextassistive

DaceyTom2:
### Description
Bug: 82988 replace sctitle with sclinktext and sclinttextassistive

List of proposed changes:
- Dashboard now uses sclinktext and sclinttextassistive
- Profile now uses sclinktext and sclinttextassistive
- Both pages use aria labels to read out assistive text

Check Dashboard and Profile for aria labels read out correctly
Content should be visible with no errors on Dashboard and Profile pages

### Additional Notes
scTitle will no longer exist in AEM after PR is merged


shewood (approved):
Looks Good.

PR #102, opened 2022-10-28
Cypress code improvements

manishasharma5757:
### Description
Cypress test cases code improvements


DaceyTom2 (approved):
Looks much cleaner! Be careful for content changes as these words may change

PR #103, opened 2022-10-31
Sch 84580 1

shewood:
### Description

List of proposed changes:
fixed the spacing as outlined in the Figma
moved the list for the profile page into its own component so that can styled more easily and also it doesn't need the heading so this will also fix the accessibility issue.

### DIFF
--- a/components/ProfileTasks.js
+++ b/components/ProfileTasks.js
@@ -45,13 +45,11 @@ export default function ProfileTasks(props) {
 
 ProfileTasks.propTypes = {
   dataCy: PropTypes.string,
-  taskListMR: PropTypes.shape({
-    title: PropTypes.string.isRequired,
-
+  taskList: PropTypes.shape({
     tasks: PropTypes.arrayOf(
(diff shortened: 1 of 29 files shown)


components/ProfileTasks.js
```diff
@@ -0,0 +1,63 @@
...
+          )
+        })}
+      </ul>
+    </>
+  )
+}
+
+ProfileTasks.propTypes = {
+  dataCy: PropTypes.string,
+  taskListMR: PropTypes.shape({
```
DaceyTom2, line 51: taskListMR can be changed to taskList.
Since props.taskList isn't being used you'll probably see errors with title being a required field

DaceyTom2 (changes requested):

components/ProfileTasks.js
shewood: Right, good catch!

DaceyTom2 (approved):
I think pulling from dev did something weird but overall it looks good!

PR #104, opened 2022-11-01
Bug: Fixed profile more less buttons

DaceyTom2:
### Description
Bug: Fixed profile more less buttons

List of proposed changes:
- profile more less buttons text now match design

Check profile page links and match with design


shewood (approved):
I tried it out and it looks good.

PR #125, opened 2022-11-09
Bug 87058 Dashboard new tab exceptions

DaceyTom2:
### Description
Bug 87058 Dashboard new tab exceptions

List of proposed changes:
- Add new tab exception array to allow for links to open in new tabs (https://www.canada.ca/en/services/benefits/ei/ei-regular-benefit/apply.html#gc-document-nav, https://www.canada.ca/en/services/benefits/ei/employment-insurance-reporting.html)
- Since opening in a new tab we bypass the exit beta pop up logic

Open the app and check if the links open in new tabs. Others should keep existing logic.

### DIFF
--- a/components/BenefitTasks.js
+++ b/components/BenefitTasks.js
@@ -2,9 +2,13 @@ import PropTypes from 'prop-types'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import Link from 'next/link'
 import { icon } from '../lib/loadIcons'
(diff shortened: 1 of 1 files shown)


CurtisUnderwood (approved):
Looks good!

nibivi77:
Check this Link out and let me know what you think.

https://developer.mozilla.org/en-US/docs/Web/API/Window/open#avoid_resorting_to_window.open

DaceyTom2:
@nibivi77 I updated the anchor tag to use target. However, it is still not great in terms of accessibility

PR #127, opened 2022-11-10
Fix styles

shewood:
### Description
Updated some styles just for the text and font on the banner to match the figma. 
Added back some styles that were lost for the task list. 
List of proposed changes:

-
-


nibivi77 (approved):
Looks much better now!

PR #128, opened 2022-11-11
87175 beta banner popup

numbap:
### Description

List of proposed changes:

- Added popup to banner
- If no link is sent from GraphQL, forward user to wikipedia as a placeholder value. 

 - Click on button "Exit Beta Version". Modal should open. 
 - Clicking on "Exit Beta Version" in modal should forward user to placeholder next page. 

### Additional Notes

"scDestinationURLEn" and "scDestinationURLFr" generate error when null. GraphQL query should be updated to return URL values.


shewood (approved):
Button now works

PR #130, opened 2022-11-14
87783 Fixed security-settings mapping

CurtisUnderwood:
### Description

List of proposed changes:

- Changes made to security-settings mapping as structure changed in AEM


nibivi77 (approved):
This is good Curtis, thank you for fixing the AEM problem!

A few things to consider for some future work. 
- we can refactor how many times we call `findFragmentByScId` down to only `2` calls
- Some of the content fetched from the query is never used. Although it is not a problem per say.

PR #132, opened 2022-11-18
Adding all AEM popup content.

nibivi77:
### Description

List of proposed changes:

- Adding all AEM content.

- To test *exit-beta-popup*
- Click exit beta button in both french and english. (Content should be the same as in aem)

image

- To test *beta-popup-page-not-available*
- Comment out this block of code in the dashboard page (Line 101):
```  
const popupContent = await getBetaPopupExitContent().catch((error) => {
    logger.error(error)
    // res.statusCode = 500
    throw error
  })
```
- Uncomment the block of code below:
```
const popupContent = await getBetaPopupNotAvailableContent().catch((error) => {
      logger.error(error)
      // res.statusCode = 500
      throw error
    })
```

image

### DIFF
--- a/__tests__/components/ExitBetaModal.test.js
+++ b/__tests__/components/ExitBetaModal.test.js
@@ -12,14 +12,24 @@ describe('Exit Beta Modal', () => {
         closeModal={() => {}}
         closeModalAria={'close'}
         continueLink="/"
+        popupId={'Test Id'}
+        popupTitle={'Test Title'}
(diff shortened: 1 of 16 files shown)


nibivi77:
> Nice work! I assume we leave the commented part until we figure out some logic for it.

Yes exactly. I don't know how we will decide which content to show in the popup. In the meantime, everything is there and ready to be used.

PR #134, opened 2022-11-21
added beta page for status and documents

shewood:
### Description
Create a mock page for leaving or staying on the beta site for Submitting documents and Status updates.

List of proposed changes:
Added images of the pages and these images will be displayed as a page.
there is only a desktop and mobile image. 
Back to dashboard button at the bottom should return the user to the dashboard page. 
Need to use the url to view the page.


DaceyTom2 (approved):
Mobile, desktop, French and English seem to be working well

PR #135, opened 2022-11-21
86051 Fixes FA icon flash for all icons in prod

CurtisUnderwood:
### Description

List of proposed changes:

- There was a small flash of large unstyled FA icons on page load in the production build. This should force them all to wait for styling.

- Shouldn't see an icon flash in the prod build


nibivi77 (approved):
One way to find out if this worked! :)

PR #137, opened 2022-11-22
Forgot other components.

nibivi77:
Quick bug fix.


shewood (approved):
I tried it and now it works.

PR #141, opened 2022-11-23
Bug fix - console error from Next Link in modal

DaceyTom2:
### Description
Bug fix - console error from Next Link in modal

List of proposed changes:
- changed next link to anchor in modal

### Additional Notes
Since the modal is in a portal it has issues with next link
It is fine to use an anchor tag since this link always navigates outside of the app, therefore, not using next's routing


nibivi77 (approved):
Beautiful! Thank you!

PR #145, opened 2022-11-30
88232 Pulling in dropdown text from AEM

CurtisUnderwood:
### Description

List of proposed changes:

- Card dropdown text now comes form AEM and is unique to dashboard cards

### DIFF
--- a/graphql/mappers/my-dashboard.js
+++ b/graphql/mappers/my-dashboard.js
@@ -15,6 +15,7 @@ export async function getMyDashboardContent() {
           return {
             id: fragment.scId,
             title: fragment.scTitleEn,
+            dropdownText: fragment.schTasks[0].scLinkTextEn,
             lists: fragment.schLists.map((list) => {
               return {
                 title: list.scTitleEn,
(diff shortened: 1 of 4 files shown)


graphql/mappers/my-dashboard.js
```diff
@@ -15,6 +15,7 @@ export async function getMyDashboardContent() {
           return {
             id: fragment.scId,
             title: fragment.scTitleEn,
+            dropdownText: fragment.schTasks[0].scLinkTextEn,
```
DaceyTom2, line 18: Would it be possible to get this by scid?

graphql/mappers/my-dashboard.js
```diff
@@ -51,6 +52,7 @@ export async function getMyDashboardContent() {
           return {
             id: fragment.scId,
             title: fragment.scTitleFr,
+            dropdownText: fragment.schTasks[0].scLinkTextFr,
```
DaceyTom2, line 55: same as above

graphql/mappers/my-dashboard.js
DaceyTom2: Approving since IDs won't match per card

CurtisUnderwood: Discussed, will need to change the IDs in AEM

DaceyTom2 (approved):
Looks good!

PR #147, opened 2022-11-30
opentelemetry/auto-instrumentations-node update to 34.0

nibivi77:
Updating opentelemetry/auto-instrumentations-node dependency to latest.


DaceyTom2 (approved):
Taking care of business!

PR #149, opened 2022-11-30
88232 Pulling in dropdown text from AEM

CurtisUnderwood:
### Description

List of proposed changes:

- Card dropdown text now comes form AEM and is unique to dashboard cards


DaceyTom2 (approved):
Looks good! Talk with Isabelle after about schTasks IDs not matching

PR #157, opened 2022-12-07
90266 create graph ql for ei contact

numbap:
### Description

List of proposed changes:

- GraphQL function that maps details for EI Contact page
-


graphql/queries/contact-employment-insurance.graphql
```diff
@@ -0,0 +1,78 @@
+{
+  schListv1ByPath(
+      _path: "/content/dam/decd-endc/content-fragments/sch/lists/contact-us/ei-contact-methods"
```
CurtisUnderwood, line 3: Looks good! But I think we'll need to direct our query towards the page eg. '/sch/pages/contact-us-pages/ei-contacts' and then pull in the contact-methods that are associated with the page fragment as part of a greater query of the page. This way we'll be able to get additional info from the page like its meta data, title, etc.

graphql/mappers/contact-employment-insurance.js
```diff
@@ -0,0 +1,100 @@
+import clientQuery from '../client'
```
CurtisUnderwood, line 1: I also think it'd be beneficial to integrate this mapper into the contact-employment-insurance page in this PR, that way we can tell if the GQL query and mapper are working the way we want them to before its merged. Should be as simple as importing this file and substituting the parseContactMethods function

package.json
```diff
@@ -31,6 +31,8 @@
     "@opentelemetry/api": "^1.2.0",
     "@opentelemetry/auto-instrumentations-node": "^0.34.0",
     "@opentelemetry/sdk-node": "^0.33.0",
+    "apollo-boost": "^0.4.9",
+    "apollo-link-http": "^1.5.17",
```
DaceyTom2, line 35: Are these being used?

DaceyTom2 (commented):
I agree with Curtis. The only other thing I notice are two added npm packages that don't appear to be used

PR #165, opened 2022-12-15
added spacing between buttons on the modal

shewood:
### Description
The spacing between the buttons on the modal was missing. 
List of proposed changes:
Added spacing between the buttons. 

-
-


DaceyTom2 (approved):
Good catch!

PR #169, opened 2022-12-19
93112 contact pagep icon

numbap:
### Description

List of proposed changes:

- Added background row highlight
- Added icon to row

### Additional Notes
 - Still need to add button capability, once variable is available in AEM.


components/contact/ContactSectionRow.js
```diff
@@ -1,18 +1,23 @@
...
+      className={`grid grid-cols-12 gap-4 border-t-2 my-4 p-2 ${
         highlight && 'bg-blue-100'
       }`}
       key={index}
     >
-      <div className="col-span-1 font-bold text-xl">{label}</div>
-      <div className="col-span-4 [&_ul]:list-inside [&_ul]:ml-4 [&_ul]:list-disc">
+      <div className="col-span-3 font-bold text-xl">{label}</div>
+      <div className="col-span-1 grid justify-center m-2">
+        {icon && <FontAwesomeIcon icon={faClock} />}
```
CurtisUnderwood, line 18: Icon looks good but is there a way to make this more dynamic? We're assuming the only icon we'll ever need is faClock which is fine just might want to clarify that this is the case with content since the field in AEM is currently for any icon.

Also we may need to pull in an icon bg-colour from AEM to avoid doing that logic on the page.

numbap: I debated this. The problem is that the icon library can be inconsistent, and it would be fairly easy for someone to submit an icon definition that's valid under FA, but still fails in the app. We would need to constrain the choices of icons that we'll accept from AEM.

shewood: Take a look at how the icons are done for the BenefitTasks and add the icon to the loadicons.js list. this way any icon can be used.

numbap: Done.

numbap (commented):
Changed prop name to resolve naming conflict

shewood (approved):
Looks good. 
Some small design tweaks will be needed at some point.

PR #171, opened 2023-01-06
96051 create modal

numbap:
### Description

List of proposed changes:

- Created CountDown and SignedOut modal.  
- Created temporary modaldemo page to trigger modals. 
-

Unit tests create a snapshot of current modal state.


shewood (commented):
Is there a way to test this without a snapshot test? They are hard to maintain.

components/sessionModals/CountDown.js
```diff
@@ -0,0 +1,101 @@
...
+        <div className="flex-none flex items-top justify-center pr-3">
+          <FontAwesomeIcon
+            icon={icon['triangle-exclamation']}
+            size="lg"
+            color="orange"
+          />
+        </div>
+        <div className="flex-auto">
+          <p className="mr-6">
+            You have been active for a while so we will sign you out to keep
```
nibivi77, line 72: I assume we mean "inactive'? and not "active"

Do we know if we will get this from AEM?

nibivi77 (commented):
`CountDown.js` and `SignedOut.js` are very similar to `ExitBetaModal.js`.

Could we find a way to reduce code duplication and combine them in a common Modal component? 

Just thoughts.. we can discuss this !

components/sessionModals/CountDown.js
numbap: Yes. The content will come from AEM.

nibivi77 (approved):

PR #183, opened 2023-01-16
96037 map base url to aem link

DaceyTom2:
### Description
Mapped base urls based off AEM property

List of proposed changes:
- MSCA links now work
- MSCA links can be configured to each environment

Check MSCA links go to expected spot. (They shouldn't point to our own application)


shewood (approved):
Tried it out and the links are working, (not the security ones though, they still are broken at the other end)

PR #186, opened 2023-01-18
96052 trigger modal after time

numbap:
### Description

List of proposed changes:

- Coundown and Signout modals trigger on dashboard page after 1 and 2 minutes respectively. 
- In order to trigger modals, expires.active variable must be set to true. Timer is deactivated by default.

### DIFF
--- a/pages/my-dashboard.js
+++ b/pages/my-dashboard.js
@@ -1,6 +1,6 @@
 import CountDown from '../components/sessionModals/CountDown'
 import SignedOut from '../components/sessionModals/SignedOut'
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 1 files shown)


pages/my-dashboard.js
```diff
@@ -45,6 +52,38 @@ export default function MyDashboard(props) {
...
+              id="CountDown"
+              deadline={expires.logout}
+              {...props.popupStaySignedIn}
+            />
+          )
+        )
+      } else return
+    }, 1000)
+    return () => clearInterval(id)
+  }, [expires])
```
DaceyTom2, line 85: This might lead to some quirky behaviour since this code is run every time the expires state is updated.
It could lead to multipl intervals being set. I'm not sure it would result in any bugs since we can only have one modal but it could flicker.

I'd recommend leaving [expires] empty [] so the code only ever runs once

DaceyTom2 (approved):

PR #196, opened 2023-01-27
Profile page changes

nibivi77:
Added changes to Profile page as per [design](https://miro.com/app/board/uXjVOjIfoTM=/?moveToWidget=3458764535881837938&cot=14)

[Azure Task](https://dev.azure.com/VP-BD/DECD/_sprints/taskboard/Secure%20Client%20Hub/DECD/Secure%20Client%20Hub/Medicine%20Hat%20%E2%80%93%20SCH%20Beta%20Sprint%2013?workitem=98989)

Removed unused cypress tests as we no longer use cards.

Worth noting:
 - Simple cypress tests can be added.


graphql/mappers/profile.js
```diff
@@ -8,51 +7,113 @@ export async function getProfileContent() {
...
           }
         })
         .filter((e) => e),
+      lookingFor: {
+        title: response.data.schPagev1ByPath.item.scFragments.find(
+          (element) => element.scId === 'looking-for-security-settings'
+        ).scContentEn.json[0].content[0].value,
+        subText: [
+          response.data.schPagev1ByPath.item.scFragments.find(
+            (element) => element.scId === 'looking-for-security-settings'
```
CurtisUnderwood, line 44: Can we define the data for certain components in a constant at the top of the page like on the security-settings page. This will keep us from running multiple find fragment calls and will probably make it easier to error proof in the future. 
ex. (we may be able to do this without separating en and fr into two calls)
```
const enLookingForFragment = findFragmentByScId(
    response,
    'looking-for-profile-settings'
  ).scContentEn
```

graphql/mappers/profile.js
```diff
@@ -8,51 +7,113 @@ export async function getProfileContent() {
...
-            title: fragment.scTitleEn,
-            lists: {
-              tasks: fragment.schTasks.map((list) => {
+      heading: response.data.schPagev1ByPath.item.scFragments.find(
+        (element) => element.scId === 'profile-intro'
+      ).scContentEn.json[0].content[0].value,
+      list: response.data.schPagev1ByPath.item.scFragments
+        .map((element) => {
+          if (
+            element.scId !== 'profile-cards' &&
```
CurtisUnderwood, line 16: We could probably reduce this to check if this is one of the three fragments ei, cpp, oas-profile we're looking for.

nibivi77: Should be good!

graphql/mappers/profile.js
nibivi77: Should be good!

DaceyTom2 (approved):
Matches design well. Thank you for cleaning up tests

PR #197, opened 2023-02-01
applied recommended changes

numbap:
### Description

List of proposed changes:

- applied changes requested by AA team
-


components/MetaData.js
```diff
@@ -14,19 +14,39 @@ export default function MetaData(props) {
...
-        <meta name="dcterms.title" content="My Service Canada Account" />
-        <meta name="dcterms.language" content="eng" />
-        <meta name="dcterms.creator" content="Service Canada" />
-        <meta name="dcterms.accessRights" content="2" />
+        <meta
+          name="dcterms.title"
+          content={
+            props.language === 'en'
+              ? 'My Service Canada Account - My dashboard'
+              : 'Mon dossier Service Canada - Mon tableau de bord'
```
DaceyTom2, line 22: This changes all the pages meta data to use "- my dashboard". I'd recommend passing in page related information using the component props

DaceyTom2 (changes requested):
Page specific values will have to come through props. It might be worth getting theses values from the en/fr locale files or AEM as well

components/MetaData.js
numbap: props.language, dcterms.creator and dcterms.service will continue to be manually coded instead of pulling from AEM.

DaceyTom2 (approved):

PR #198, opened 2023-02-02
Bug fix: App build error on static props

DaceyTom2:
Bug fix: App build error on static props

### Description
Bug fix: App build error on static props

List of proposed changes:
- changed static props to server side props

Ensure app builds


shewood (approved):
The build is working in TC with this change.

PR #200, opened 2023-02-02
Aa script update

numbap:
### Description

List of proposed changes:

- mapped AA variables in header as per requirements, using page props to guide metadata. 
-


nibivi77 (approved):
Great job as always! :+1:

PR #208, opened 2023-02-06
fix build error with getServerSideProps

shewood:
### Description
We need to use getServerSideProps otherwise the build will fail in TC. 
List of proposed changes:
Changed getStaticProps to getServerSideProps.
-
-


CurtisUnderwood (changes requested):
We'll also need getServerSideProps on _error.js

CurtisUnderwood (approved):
LGTM

PR #209, opened 2023-02-06
99714 client routing

numbap:
## [ADO-99714](https://dev.azure.com/VP-BD/DECD/_workitems/edit/99714)
Fixed [AB#99714](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/99714)

### Description

List of proposed changes:

- Used beta DS to add Link to menu
- Testing build. Please do not merge yet.


nibivi77 (commented):
Seems to function properly. The only thing I noticed is when you navigate to another page, the drop-down menu stays open.

image

I suspect that this comes from Design System. Do we care about this?

DaceyTom2 (dismissed):
Seems to work well. The first time I clicked on the menu it wasn't changing the page. Worth keeping a mental note to check from time to time

shewood (changes requested):
to skip an entire test you just add .skip like this
it.skip('Validate da....

components/Layout.js
```diff
@@ -25,7 +26,11 @@ export default function Layout(props) {
 
   return (
     <>
-      <MetaData language={props.locale} data={props.meta}></MetaData>
+      <MetaData
+        language={props.locale}
+        data={props.meta}
+        lang={!props.locale ? 'en' : props.locale}
```
nibivi77, line 32: I think `lang` prop can be removed since its not being used in <MetaData> and that `language` is doing the same thing.

nibivi77 (changes requested):

nibivi77 (approved):
Awesome!

PR #213, opened 2023-02-08
fixed blue highlight

shewood:
### Description
Updated tailwind typography
used prose to style so that the spacing outlined from AEM is used. This also solves some accessibility issues. 
Fixed the spacing of the blue highlight in the contact us pages. 

### Additional Notes
Note that there is something still janky with the contact us pages in the way they are rendered but that will be a new task.

### DIFF
--- a/components/contact/ContactSection.js
+++ b/components/contact/ContactSection.js
@@ -10,16 +10,16 @@ export default function ContactSection({ title, intro, id, details }) {
       id={id}
       data-testid="contactSection-test"
     >
-      <h2 className="py-2 text-32px md:py-6 md:text-4xl font-display font-bold">
+      <h2 className="py-2 text-32px md:pt-6 md:text-4xl font-display font-bold">
         {title}
       </h2>
       <div
-        className="list-disc [&_ul]:list-outside [&_ul]:pl-4 [&_ul]:ml-4 [&_ul]:list-disc font-body text-xl"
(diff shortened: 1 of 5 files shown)


numbap (approved):
Looks good

components/contact/ContactSection.js
```diff
@@ -10,16 +10,16 @@ export default function ContactSection({ title, intro, id, details }) {
...
       </h2>
       <div
-        className="list-disc [&_ul]:list-outside [&_ul]:pl-4 [&_ul]:ml-4 [&_ul]:list-disc font-body text-xl"
+        className="pb-4 prose max-w-none prose-p:text-xl prose-p:mb-2 prose-p:font-body prose-ul:my-0 prose-ul:ml-2 prose-li:font-body prose-li:text-xl prose-li:marker:text-black"
         data-cy="section1"
       >
         <Markdown>{intro}</Markdown>
       </div>
-      <div data-cy="section2">
+      <h3 className=" border-y-2 divide-y-2 " data-cy="section2">
```
DaceyTom2, line 22: I don't believe wrapping a section in a H3 is a good practice. The H tag should be to identify a heading. Perhaps a span or list element would suit better?

DaceyTom2 (changes requested):

components/contact/ContactSection.js
shewood: I can leave it as a div in this case then.

PR #215, opened 2023-02-08
102099 Added GH Action for changelog generation

CurtisUnderwood:
[ADO-102099](https://dev.azure.com/VP-BD/DECD/_workitems/edit/102099)
Description
List of proposed changes:

Added Release Please v. 3.1.2
On push to main (when our dev branch is merged with release branch) Release Please automatically creates a "release" PR that will need to be merged. This PR will automatically generate a Release in GitHub equipped with a changelog based on our commits. Commits must follow https://www.conventionalcommits.org/en/v1.0.0/ Conventional Commits Guidelines.

What to test for/How to test
Can be tested on any push to main

Additional Notes
https://www.conventionalcommits.org/en/v1.0.0/

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,13 @@
+on:
+  push:
+    branches:
+      - main
+name: release-please
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3.1.2
+        with:
+          release-type: node
+          package-name: release-please-action
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,8 @@
         "prop-types": "^15.8.1",
         "react": "17.0.2",
         "react-dom": "17.0.2",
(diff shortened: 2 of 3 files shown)


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,13 @@
+on:
+  push:
+    branches:
+      - main
+name: release-please
```
shewood, line 5: would it be ok to call it 'change log' instead of 'release-please' so that we can tell what it is quickly in the actions list?

shewood: Although maybe it's self explanatory enough? Thoughts? 
image

CurtisUnderwood: Happy to change it to change-log I think release-please might be more encompassing though because the action covers publishing Release versions on GitHub which also include change-logs.

shewood: I am fine with that too.

shewood (approved):
Nice, I see this action in the list on github

PR #219, opened 2023-02-10
Revert DS error pages.

nibivi77:
Revert error pages since DS pages would break our builds.


shewood (approved):
Although not ideal, this is now building in TC.

PR #221, opened 2023-02-14
103312

shewood:
### Description

List of proposed changes:
Added an id to the button on the contact us pages for accessibility. 
Updated the styling slightly so that the markdown from AEM can be used.


components/contact/ContactProvinceRow.js
```diff
@@ -16,7 +16,9 @@ const ContactProvinceRow = ({ label, items, id }) => {
         >
           {items.map((x, i) => (
             <div className="col-span-2 md:col-span-1 py-3" key={i}>
-              <p className="font-bold font-display">{x && x.content}</p>
+              <div className="prose prose-p:text-xl prose-p:font-display font-bold">
+                <Markdown>{`${ap(x.content, ' ')}`}</Markdown>
+              </div>
```
nibivi77, line 21: Just a small change. We can remove the surrounding `<div>` and simply have this:

`<Markdown className="prose prose-p:text-xl prose-p:font-display font-bold">{`${ap(x.content, ' ')}`}</Markdown>`

shewood: I can make it a span if that's preferable but it's holding the styling so it's needed.

nibivi77 (approved):
:)

PR #227, opened 2023-02-16
Update link in banner to open in new tab

krischarbonneau:
### Description
This PR updates the link in the banner to open in a new tab. In doing so, it also became necessary to add sr-only text to alert screen readers that the link is external. To do this, I've added a new query for the dictionary, and filter through it for the external link text in the opt out banner mapper. The dictionary can potentially also be used for other terms in the future.

Fixes [AB#107391](https://dev.azure.com/VP-BD/DECD/_workitems/edit/107391)

1. Pull in branch
2. Type `npm run dev`
3. You can either fire up a screen reader and navigate to the link in the banner to hear it say `opens in a new tab` or remove the `sr-only` class to see that the text is present

### DIFF
--- a/graphql/mappers/beta-banner-opt-out.js
+++ b/graphql/mappers/beta-banner-opt-out.js
@@ -2,9 +2,12 @@ import clientQuery from '../client'
 
 export async function getBetaBannerContent() {
   const queryOptOut = require('../queries/beta-banner-opt-out.graphql')
+  const queryDictionary = require('../queries/dictionary.graphql')
   const resOptOut = await clientQuery(queryOptOut)
+  const resDictionary = await clientQuery(queryDictionary)
 
   const resOptOutContent = resOptOut.data.schContentv1ByPath.item || {}
+  const resDictionaryContent = resDictionary.data.dictionaryV1List.items || {}
 
   const mappedBanner = {
     en: {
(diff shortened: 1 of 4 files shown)


graphql/mappers/beta-banner-opt-out.js
```diff
@@ -13,6 +16,9 @@ export async function getBetaBannerContent() {
       bannerLink: resOptOutContent.scFragments[0].scLinkTextEn,
       bannerLinkHref: resOptOutContent.scFragments[0].scDestinationURLEn,
       bannerButtonText: resOptOutContent.scFragments[1].scLinkTextEn,
+      bannerButtonExternalText: resDictionaryContent.filter(
+        (entry) => entry.scId === 'opens-in-a-new-tab'
+      )[0].scTermEn,
```
DaceyTom2, line 21: resDictionaryContent.find would be faster and easier to read

graphql/mappers/beta-banner-opt-out.js
```diff
@@ -2,9 +2,12 @@ import clientQuery from '../client'
 
 export async function getBetaBannerContent() {
   const queryOptOut = require('../queries/beta-banner-opt-out.graphql')
+  const queryDictionary = require('../queries/dictionary.graphql')
   const resOptOut = await clientQuery(queryOptOut)
+  const resDictionary = await clientQuery(queryDictionary)
 
   const resOptOutContent = resOptOut.data.schContentv1ByPath.item || {}
+  const resDictionaryContent = resDictionary.data.dictionaryV1List.items || {}
```
DaceyTom2, line 10: There is a way to query once and pass around a reference.
I'm not positive but I believe custom hooks handle this issue for generic content

DaceyTom2 (changes requested):

graphql/mappers/beta-banner-opt-out.js
krischarbonneau: As per our conversation, something to revisit. Currently we're only using the dictionary for the one term, but the opportunity is there to use it more.

graphql/mappers/beta-banner-opt-out.js
krischarbonneau: Fixed :smiley:

PR #233, opened 2023-02-17
fixed e2e tests

numbap:
## [ADO-102876](https://dev.azure.com/VP-BD/DECD/_workitems/edit/102876)
fixed [AB#102876](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/102876)
### Description

List of proposed changes:

- Fixed error where invalid HTML was being generated as a result of DIV inside of BUTTON tag. 
-

### DIFF
--- a/components/ViewMoreLessButton.js
+++ b/components/ViewMoreLessButton.js
@@ -4,7 +4,7 @@ import { solid } from '@fortawesome/fontawesome-svg-core/import.macro'
 
 export default function ViewMoreLessButton(props) {
   return (
-    <div
+    <button
       className={
         'text-xl leading-8 text-deep-blue-60d hover:text-blue-hover ' +
         props.className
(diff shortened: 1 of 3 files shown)


components/ViewMoreLessButton.js
```diff
@@ -4,7 +4,7 @@ import { solid } from '@fortawesome/fontawesome-svg-core/import.macro'
 
 export default function ViewMoreLessButton(props) {
   return (
-    <button
+    <div
```
DaceyTom2, line 7: Changing buttons to divs breaks accessibility. Can this be investigated further?

DaceyTom2 (changes requested):

components/ViewMoreLessButton.js
krischarbonneau: Could we potentially use the details & summary component here instead?

shewood (changes requested):
Looking at the link in the task, this is coming from the design system for the navigation menu. 
image

PR #236, opened 2023-02-20
Update metadata title orientation

krischarbonneau:
### Description
Fixes [AB#107587](https://dev.azure.com/VP-BD/DECD/_workitems/edit/107587)

In this PR, I've updated the title tag for each page so the page name comes first, followed by `My Service Canada Account`. The exception to this is any page that currently has the title as `My Service Canada Account - Canada.ca` as having the Canada.ca first doesn't really make any sense, nor does it look right.

1. Pull in branch
2. Type `npm run dev`
3. Navigate through the site and ensure each page's title tag has been updated to reflect the format explained above

### Additional Notes
Some of the beta pages didn't have unique metadata, will we be implementing that? Are we eventually going to migrate all of this to AEM?

Currently we don't have a test written for the `contact-us` page, only the child pages. This should be addressed in another PR.

### DIFF
--- a/__tests__/pages/contact-canada-pension-plan.test.js
+++ b/__tests__/pages/contact-canada-pension-plan.test.js
@@ -141,19 +141,19 @@ describe('CPP Contact Us Page', () => {
         locale: 'en',
         meta: {
           data_en: {
+            title: 'Contact Canada Pension Plan - My Service Canada Account',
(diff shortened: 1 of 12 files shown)


DaceyTom2 (approved):
Lg2m

PR #238, opened 2023-02-21
New error pages

nibivi77:
### Description

Trying to add error pages from DS and see if they break our builds.

### DIFF
--- a/pages/404.js
+++ b/pages/404.js
@@ -22,7 +22,7 @@ export async function getStaticProps({ locale }) {
   /* Place-holder Meta Data Props */
   const meta = {
     data_en: {
-      title: 'My Service Canada Account - 404',
+      title: '404 - My Service Canada Account',
       desc: 'English',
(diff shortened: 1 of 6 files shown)


pages/404.js
```diff
@@ -0,0 +1,53 @@
...
+  )
+}
+
+export async function getStaticProps({ locale }) {
+  const langToggleLink = locale === 'en' ? '/fr/404' : '/404'
+
+  /* Place-holder Meta Data Props */
+  const meta = {
+    data_en: {
+      title: 'My Service Canada Account - 404',
```
krischarbonneau, line 25: as per [AB#107587](https://dev.azure.com/VP-BD/DECD/_workitems/edit/107587), this should be `404 - My Service Canada Account, same thing for FR and the other error page

nibivi77: Should be good now.

krischarbonneau (approved):

PR #247, opened 2023-03-03
Fixed FOUC issue on initial page load

krischarbonneau:
### Description
Fixes [AB#98971](https://dev.azure.com/VP-BD/DECD/_workitems/edit/98971)

This PR fixes the FUOC (flash of unstyled content) issue that our application was having whenever navigating to a new page. This is achieved by adding a custom `_document.js` with a dummy script directly after the body tag before the main content is loaded so the page is only rendered once the CSS is loaded.

This is a common issue that you can read more on [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1404468) and [here](https://github.com/vercel/next.js/issues/15642)

1. Pull in branch
2. Type `npm run build` and `npm run start`
3. Navigate to the dashboard and from the menu, click any of the links and ensure that there isn't any flash of unstyled content
4. To confirm that this fix is working, comment out the script tag in `_document.js` and repeat the steps above. You should notice unstyled content flashing.

### DIFF
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -1,0 +1,15 @@
+import { Html, Head, Main, NextScript } from 'next/document'
+
+export default function Document() {
+  return (
+    <Html>
+      <Head />
+      <body>
+        {/* Dummy script that will force the page to render only after CSS is loaded */}
+        <script>0</script>
+        <Main />
+        <NextScript />
+      </body>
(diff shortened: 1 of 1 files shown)


numbap (commented):
Looks good.

numbap (approved):

PR #254, opened 2023-03-20
Removing beta files.

nibivi77:
Removing beta files. 

### Note
- `SWAP_USER_TESTING_LINKS = "false"` can be removed from your .env file.


nibivi77:
A commit was not signed. Opening a new PR.

PR #257, opened 2023-03-27
114208 track non ds links

numbap:
## [ADO-114208](https://dev.azure.com/VP-BD/DECD/_workitems/edit/114208)

fixed [AB#114208](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/114208)

### Description

List of proposed changes:

- Added AA tracking to non-DS link clicks
- Does not track open-close actions

 - Look at all links that are not routed through DS. There should be a prop on the HTML component in the style of "data-gc-analytics-customclick"


krischarbonneau (approved):
Looks good!

PR #272, opened 2023-04-19
Fix logout functionality

krischarbonneau:
### Description
Fixes [AB#115882](https://dev.azure.com/VP-BD/DECD/_workitems/edit/115882)

This PR adds logout functionality to the sign out button in the menu and the timeout modal on the `my-dashboard` page. The current global logout from ECAS doesn't redirect back upon logging out, so once logged out you'd have to open a new tab to log in again. This functionality is implemented by calling the `signOut` event from `next-auth` when the `Sign out` menu item is clicked, which ends the `next-auth` session, and then redirecting to `/auth/logout` (by default the `callbackURL` option from the `signOut` event only works with same origin URLs, hence the need to redirect to the logout page and redirect to an external link from there). In `/auth/logout`, the logout URL is obtained in `getServerSideProps` from `/lib/auth.js`, where I've added a new function to return a URL with the client ID and shared session ID. Upon loading the page, a user will see the loading spinner (for now) until they are redirected to the global log out page.

I've also changed the name and location of the `signedout` page to `/auth/login` to match the logout page (and because the naming of the page was confusing).

1. Pull in branch
2. Type `npm i`
3. Either go to the auth-dev branch which has been deployed with these changes **OR** fire up the proxy and continue with the below steps
4. Type `npm run dev`
5. Upon selecting a language, navigate through the login flow until you end back at `/my-dashboard`
6. Select `Sign out` from the menu
7. You should be redirected to `/auth/logout` and then to the global logout page


components/Layout.js
```diff
@@ -8,7 +8,7 @@ import MetaData from './MetaData'
 import { signOut } from 'next-auth/react'
 import PhaseBanner from './PhaseBanner'
 import Modal from 'react-modal'
-import { useEffect } from 'react'
+import { useEffect, useCallback } from 'react'
```
numbap, line 11: Added useCallback, but never used.

pages/auth/logout.js
```diff
@@ -0,0 +1,64 @@
...
+    (await getLogoutURL(req).catch((error) => {
+      logger.error(error)
+      res.statusCode = 500
+      throw error
+    }))
+
+  return {
+    props: {
+      locale,
+      logoutURL: logoutURL || null,
```
numbap, line 61: The null value causes the page to crash, and setting it to "" causes an infinite loop. Is there some way to elegantly fail?

numbap (commented):
There is an AA bug due to the fact that AA is attempting to pull an insecure URL, which violates the security policy in next.config. Will address with PP team.

components/Layout.js
krischarbonneau: Was part of some previous testing. Missed this, thanks.

pages/auth/logout.js
krischarbonneau: This should never happen if there is a valid auth session, but in the case it does fail, I've updated it so it redirects to the index page instead.

numbap (approved):

PR #280, opened 2023-05-03
121071 middleware fix

krischarbonneau:
### Description
This PR updates the middleware to prevent users from having a selected language on the index page and will redirect them to `/` if they try to navigate there (either manually or via signing out in local dev).

This PR also removes the router instance in `/auth/logout` since it was causing some weird behavior in favor of using `window.location.replace()` instead.

Fixes [AB#121071](https://dev.azure.com/VP-BD/DECD/_workitems/edit/121071)

I've also updated the logoutURL prop to be more fail safe.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/my-dashboard` and select "Sign out" from the menu. You should be redirected back to `/`
4. Repeat step 3 but sign out via the modal

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -4,15 +4,22 @@ import { NextRequest, NextResponse } from 'next/server'
 const PUBLIC_FILE = /\.(.*)$/
 
 export async function middleware(req) {
+  const { nextUrl, url } = req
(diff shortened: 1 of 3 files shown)


numbap (approved):
Looks good.

PR #292, opened 2023-05-31
Update modal timeout values and added anchors to privacy page

krischarbonneau:
### Description
This PR updates the modal timeout values temporarily until we get more direction. Now, a user will be prompted after 5 minutes of inactivity, and redirected after an additional 15 minutes. This PR also adjusts the privacy link in the footer to link to our internal page. I've split the markdown coming from AEM using a regular expression and have attached anchors to the privacy section and terms and condition section (of which I'm only currently linking to the terms and conditions since linking to the privacy section would put the user part way down the page when that section is already at the top which is bad UX, but I've kept that id there for consistency).

1. Pull in branch
2. Type `npm run dev`
3. Scroll down to the bottom of the page and select both links in the footer and confirm you are redirected to the terms and conditions section when selecting that link, and to the top of the page when selecting the other
4. Confirm you are not redirected to the index page after 1 minute and 30 seconds of inactivity like before

### Additional Notes
This is a temporary fix given MSCA should have these 2 pages separate. TBS has been made aware and they have a backlog item created to address this issue.

### DIFF
--- a/components/MultiModal.js
+++ b/components/MultiModal.js
@@ -27,8 +27,8 @@ export default function MultiModal(props) {
 
   const { isPrompted, reset, getRemainingTime } = useIdleTimer({
     onIdle: handleOnIdle,
-    promptBeforeIdle: 0.5 * 60 * 1000, // 30 seconds
-    timeout: 1 * 60 * 1000, // 1 minute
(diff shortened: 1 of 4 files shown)


shewood (approved):
I tried it out, looks good.

PR #297, opened 2023-06-16
Fix logout functionality

krischarbonneau:
### Description
This PR removes the `signOut` event call from the DS menu component and implements it asynchronously on the logout page. This was an oversight before as the session would not end if the user were to sign out from the time out modal. This PR also updates `next-auth` to the latest version. 

1. Pull in branch
2. Type `npm run dev` or `npm run build` and `npm run start`
3. Ensure you are connected to the proxy & go through the login process
4. Once on the dashboard, select sign out from the menu.
5. Confirm that you are first redirected to `/auth/logout` and then to the global sign out page
6. Ensure that the next auth session token has been wiped from your session storage (if this has been done, you should be prompted to go through the login flow again once you select a language).

### DIFF
--- a/components/Layout.js
+++ b/components/Layout.js
@@ -6,7 +6,6 @@ import {
 } from '@dts-stn/service-canada-design-system'
 import { useState, cloneElement } from 'react'
 import MetaData from './MetaData'
-import { signOut } from 'next-auth/react'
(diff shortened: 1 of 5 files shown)


nibivi77 (approved):
Looks good to me Kris! Maybe we should consider doing something if there's an error. In the catch clause.

PR #300, opened 2023-06-22
128802 language neutral aa tagging

numbap:
### Description

List of proposed changes:

- Enable language-neutral tracking for AA
-

Inspect every AA-tracked link on each page in the browser console. (Identified by data-gc-analytics-customclick prop) They should be identical in the French and English versions of the page. 

### Additional Notes

Language-neutral tracking will be broken on components that rely on DS for the AA tag.


krischarbonneau (approved):
Looks good!

PR #301, opened 2023-06-23
Add cypress tests for Login/Logout pages

krischarbonneau:
### Description
This PR adds simple cypress tests for the login and logout pages. As we cannot easily test our authentication flow through Cypress at this time due to the flow being behind firewall, these tests only test if the pages are rendering and tests that they properly redirect in the event that auth is disabled.

1. Pull in branch
2. Type `npm run e2e`
3. Run the tests for Login and Logout and confirm there are no errors


shewood (commented):
Looks good. 
No need to use const in Cypress for this case unless you are storing and comparing values.
you can just write it as 
 cy.get('[data-cy="loading-spinner"]').should('be.visible').chainshere...

numbap (approved):
Looks good.

PR #303, opened 2023-06-28
Wrap continue button in form instead of anchor

krischarbonneau:
### Description
This PR fixes an a11y issue where a button was wrapped in an anchor tag. This causes an issues for screen reader users since it will announce the anchor tag first in focus order and then the button. Placing the button in a form instead and setting the action to the URL fixes this issue. The caveat is that as soon as we set the button type to anything (in this case submit), Tailwind's preflight styling sets the background for buttons with a set type to transparent. Explicitly setting the background color again fixes this issue.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to dashboard and click any of the links, which will open the exit beta prompt. Ensure it links to another page as expected (won't actually load if not authenticated but should redirect regardless)

### DIFF
--- a/components/ExitBeta.js
+++ b/components/ExitBeta.js
@@ -39,15 +39,15 @@ export default function ExitBeta(props) {
           onClick={props.closeModal}
           text={props.popupSecondaryBtn.text}
         />
-        {/* Using anchor tag due to ref errors
-        continueLink is always external so NextJS routing can be ignored */}
-        <a
-          href={props.continueLink}
-          id={props.popupPrimaryBtn.id}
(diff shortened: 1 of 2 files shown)


krischarbonneau:
Closing this as it's not necessary. Was going off the picture in the task from the a11y audit where we were still wrapping a `button` with an anchor tag, but that is no longer the case. We should be requesting the DS team to allow for a link to be styled like a button though. Currently when we supply an href to the `Button` component it doesn't take the primary styling, so we're left with essentially a basic anchor tag with link styling.

PR #305, opened 2023-06-28
Update Card component to not use underline styling

krischarbonneau:
### Description
This PR addresses one of the suggestions made on our a11y audit. Note that this is a **usability** issue and not an accessibility issue. I've removed the underline styling from the collapsible card component as that styling is reserved for links and can be confusing to users. In addition, I found while testing that the focus for the `ViewMoreLessButton` component wasn't spanning the entirety of the parent component, which looked really bad. I updated the padding a bit and add the `w-full` class so the focus spans the entirety of the parent now. I've also added focus styling since we only had hover styling.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to the dashboard and tab through the cards, ensuring the focus spans the whole card width and that there is no underline sytling

### DIFF
--- a/components/Card.js
+++ b/components/Card.js
@@ -7,7 +7,7 @@ export default function Card(props) {
 
   return (
     <div className="border rounded border-gray-300 shadow my-6" data-cy="cards">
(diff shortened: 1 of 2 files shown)


nibivi77 (approved):
:)

PR #311, opened 2023-06-28
Update privacy page heading heirarchy

krischarbonneau:
### Description
This PR fixes an accessibility issue with our privacy page. All of our section headings were H1s, which is not allowed as there should only ever be one H1 on a page. An update has been made in AEM to change the section headings to H2s, this PR updates the Markdown styling to target the H2s instead of H1s. 

1. Pull in branch
2. Type `npm run dev`
3. Navigate to the privacy page, inspect the heading elements and ensure only the very top one is an H1 and the rest are H2s


nibivi77 (approved):
Yep that works!

PR #312, opened 2023-06-28
Updated modal to use contentLabel depending on context

krischarbonneau:
### Description
This PR updates the `MultiModal` component to set the `contentLabel` variable (which React-Modal parses into `aria-label`) based on the modal context. I've also removed the `SessionTimeout` component as it was an unused duplicate/old version of `CountDown`.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to the dashboard, click one of items in the cards to open the exit beta modal
4. Right click, inspect the element and look for `<div class="ReactModal__Content ReactModal__Content--after-open...`
5. Ensure it has an aria-label


nibivi77 (approved):
Lovely

PR #314, opened 2023-06-30
Changing wrapper strong to div.

nibivi77:
## [a11y task](https://dev.azure.com/VP-BD/DECD/_workitems/edit/129448)

Changed a `<strong>` for a `<div>` since a strong cannot have a block element within it.

### DIFF
--- a/components/contact/ContactProvinceRow.js
+++ b/components/contact/ContactProvinceRow.js
@@ -16,9 +16,9 @@ const ContactProvinceRow = ({ label, items, id }) => {
         >
           {items.map((x, i) => (
             <div className="col-span-2 md:col-span-1 py-3" key={i}>
-              <strong className="prose prose-strong:text-xl prose-strong:font-display prose-p:text-xl prose-p:font-display">
+              <div className="prose prose-strong:text-xl prose-strong:font-display prose-p:text-xl prose-p:font-display">
                 <Markdown>{`${ap(x.content, ' ')}`}</Markdown>
-              </strong>
+              </div>
               {x && (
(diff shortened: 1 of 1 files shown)


components/contact/ContactProvinceRow.js
```diff
@@ -16,9 +16,9 @@ const ContactProvinceRow = ({ label, items, id }) => {
         >
           {items.map((x, i) => (
             <div className="col-span-2 md:col-span-1 py-3" key={i}>
-              <strong className="prose prose-strong:text-xl prose-strong:font-display prose-p:text-xl prose-p:font-display">
+              <div className="prose prose-strong:text-xl prose-strong:font-display prose-p:text-xl prose-p:font-display">
                 <Markdown>{`${ap(x.content, ' ')}`}</Markdown>
-              </strong>
+              </div>
```
krischarbonneau, line 21: While changing this to a div is fine, it would require more work on the AEM side, else we lose the styling which I don't think we want given we want to apply importance to the bold content in this case. 

Rather than change it, we can instead force `markdown-to-jsx` to parse the content as inline (making it a span), which would be fine since we then have an inline element inside of another inline element instead of a block element inside of an inline element.
```suggestion
              <strong className="text-xl font-display">
                <Markdown options={{ forceInline: true }}>{`${ap(
                  x.content,
                  ' '
                )}`}</Markdown>
              </strong>
```

krischarbonneau (changes requested):

nibivi77:
This has been resolved from within AEM.

PR #323, opened 2023-07-11
fix: Page section missing lang attribute

nibivi77:
### Description

The language toggle was missing the lang attribute.
image

Problem was on the DS side. Shawn and I communicated and fixed the problem. Solved with a pre-release of Design System.


krischarbonneau (approved):
Good job investigating this :)

PR #325, opened 2023-07-11
Update opentelemetry dependencies

krischarbonneau:
### Description
This PR updates the opentelemetry dependencies to latest. Snyk tried to update sdk-node in #309 but the order of how these deps are upgraded matters for compatibility reasons (`auto-instrumentations-node` first, `sdk-node` second)

### What to test for/How to test
1. Pull in branch
2. Type `npm i`
3. Ensure application runs as expected

PR #328, opened 2023-07-11
Update package lock to fix build error

krischarbonneau:
### Description
Something happened on #325 when updating from dev that caused some necessary lines in our package lock to be nuked. Our dockerfile runs `npm ci` which fails and makes our TC build fail because of this. This PR just runs `npm i` to update the lock file to be in line with our `package.json`

1. Pull in branch
2. Type `npm i`
3. Type `npm run dev`
4. Ensure everything works

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4688,6 +4688,17 @@
         "node": ">=8.0.0"
       }
     },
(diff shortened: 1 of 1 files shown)


shewood (approved):
Looks good

PR #354, opened 2023-08-10
updated the logout e2e tests

shewood:
## [ADO-135922](https://dev.azure.com/VP-BD/DECD/_workitems/edit/135922)

Fixed [AB#135922](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/135922)

Sometimes the loading spinner was disappearing before the test could see it. This was causing the e2e test to fail randomly.
Changed the tests to check the logout from the menu.
Found a solution that hopefully will prevent the response from continuing until after the spinner has been found, then verify that it continues and logs out as expected. 

e2e test don't fail randomly

### DIFF
--- a/cypress/e2e/ContactUs.cy.js
+++ b/cypress/e2e/ContactUs.cy.js
@@ -8,12 +8,11 @@ beforeEach(() => {
   cy.visit('/contact-us')
 })
 
-it('Contact us has no detectable a11y violations on load', () => {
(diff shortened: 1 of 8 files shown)


krischarbonneau (approved):
This is great! Love how fleshed out the test is.

PR #356, opened 2023-08-22
Optimize fonts

krischarbonneau:
## [ADO-127035](https://dev.azure.com/VP-BD/DECD/_workitems/edit/127035)

Fixed [AB#127035](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/127035) (hopefully)

This PR seeks to optimize our fonts by utilizing `next/font`. I've gone off the existing implementation on Senior Journeys as a base however I've also had to change how we're loading our icons from FontAwesome due to the fact that we are no longer using babel because `next/font` requires us to be using SWC (Next.js' Rust based compiler) and that means we can't use babel macros to dynamically import the icons. This will hopefully also help to mitigate some of the issues we've had with FOUC/FOUT as well.

1. Pull in branch
2. Type `rm -rf node_modules/` and then `npm install`
3. Type `npm run dev`
4. Ensure the application runs as expected

### Additional Notes
Are we using all of the icons within `loadIcons.js` currently? I feel like we could clean it up pretty drastically.


nibivi77 (approved):
This is clean Kris. Amazing work honestly!

numbap (commented):
Looks good.

PR #370, opened 2023-09-13
Swapping Docker to standalone, re-adding Middleware, Altering ENV

lhr-git:
## [ADO-139305](https://dev.azure.com/VP-BD/DECD/_workitems/edit/139305)
Get the docker container working with middleware, and as a standalone.

- Switched NextJS production to use standalone mode (Dockerfile, next.config.js)
- Re-added middleware (middleware.ts)
- Added some things to the .gitignore and .dockerignore
- Altered the http_proxy ENV var to be http_proxy_agent (proxyconf/client.js). Reasoning: http_proxy is a reserved ENV in linux. It proxies all http traffic via it, not just our application's traffic.

build the docker image and test. Everything should be the same...

### Additional Notes
If your docker container is giving you some /graphql or auth errors, it could be the .dockerignore file's .env line. We shouldn't be including .env.... but I am still working on getting it to work in docker-compose without the .env file loading into the container as part of the first step of the build.


middleware.ts
```diff
@@ -0,0 +1,27 @@
...
+    return NextResponse.redirect(new URL(`/en${pathname}`, url))
+  }
+
+  //Redirect for index page as it's meant to be bilingual so we don't want users navigating to /en or /fr
+  if ((locale === 'en' || locale === 'fr') && pathname === '/') {
+    return NextResponse.redirect(new URL(`/`, url))
+  }
+
+  return NextResponse.next()
+}
```
krischarbonneau, line 27: This all looks good to me but if we're adding this part back we can remove the redirect rules in our next config :slightly_smiling_face:

lhr-git: Changes made, good catch.

krischarbonneau (approved):
LGTM!

PR #381, opened 2023-09-21
Sch 140369 toc

shewood:
## [ADO-140369](https://dev.azure.com/VP-BD/DECD/_workitems/edit/140369)

Fixed [AB#140369](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/140369) ,#140367

Adds the table contents component and the link component

go to security settings and the contact us pages to see the links and the table.


components/TableContents.js
```diff
@@ -0,0 +1,54 @@
...
+        <ul className="pl-[22px] ml-3.5">
+          {sectionList.map((option, index) => {
+            return (
+              <li
+                key={index}
+                className={
+                  'underline underline-offset-4 text-20px text-gray-darker underline-offset-3.1 decoration-1 pr-1 list-outside list-disc'
+                }
+              >
+                <Links
```
krischarbonneau, line 23: This can definitely use next/link instead since you can now apply styling to next/link as it just passes it down to the child `a` element

shewood: Good Idea.

components/TableContents.js
```diff
@@ -0,0 +1,54 @@
...
+                  href={option.link}
+                  id={`tableLink${index + 1}`}
+                  text={option.name}
+                  linkStyle="cardActionLink"
+                />
```
krischarbonneau, line 28: We aren't using a large majority of this file so I would just do away with it since we can just sub in the most important stuff like so:

```suggestion
                <Link
                  className="text-deep-blue-dark font-body text-xl hover:text-blue-hover focus:text-blue-hover"
                  href={option.link}
                  id={`tableLink${index + 1}`}
                  data-cy={`tableLink${index + 1}`}
                  aria-label={option.name}
                >
                {option.name}
                </Link>
```
I don't see the need to keep this component in since I moved the ternary directly into the Language component. The only reason to keep this would be for easy access to styling but I don't think that's a good enough reason to do so.

shewood: moved link to local file instead.

pages/contact-us/[id].js
```diff
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types'
-import { TableContent } from '@dts-stn/service-canada-design-system'
+// import { TableContent } from '@dts-stn/service-canada-design-system'
```
Comment (nibivi77) [inline pages/contact-us/[id].js:2]: Quick comment to remove.

nibivi77 (approved):
I like this work. Well done!

pages/contact-us/[id].js
Comment (shewood) [reply pages/contact-us/[id].js]: Good catch.

PR #382, opened 2023-09-21
Replace Header & Footer

krischarbonneau:
## [ADO-140361](https://dev.azure.com/VP-BD/DECD/_workitems/edit/140361)

Fixed [AB#140361](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/140361)

### Changelog - "fix:" for bug fixes, "feat:" for features. Read more about Conventional Commits at https://www.conventionalcommits.org/en/v1.0.0/#summary
feat:Updated app to use local Header implementation
feat:Updated app to use local Footer implementation
feat:Updated app to use local TopNav, Menu, Language switcher, Breadcrumb implementation

This PR updates our app to use a local version of the Header (and all the components it encompasses) and Footer instead of pulling from DS. While the original code is still from DS, I've taken out code that doesn't pertain to our project in order to keep things clean. I've also replaced instances of `Image` and `Link` with their corresponding Next.js components.

1. Pull in branch
2. Type `npm run dev`
3. Navigate through the app and ensure it looks and acts as expected

### Additional Notes
The tests for layout have been skipped in this PR since they are broken at a fundamental level and need to be rewritten. This will be done in a separate PR.


tailwind.config.js
```diff
@@ -44,6 +44,8 @@ module.exports = {
           '30a': '#EAEBED',
           '40': '#DDDDDD',
           '50a': '#CFD1D5',
+          'blue': '#EBF2FC',
```
shewood, line 47: would you be able to move this to the bright-blue section with maybe something like pale: #hex? 
so it would be added as bright-blue-pale or whatever.

tailwind.config.js
```diff
@@ -44,6 +44,8 @@ module.exports = {
...
+          'subFooter': '#F5F5F5',
```
shewood, line 48: this is a gray but very light so maybe a number like '10' or 'lightest'? instead of subfooter as it doesn't describe the colour in any way.

shewood (commented):
Looks good. I did notice that the menu stays open after you select an item. 
And the placement of the menu is slightly different on md

components/Footer.js
```diff
@@ -0,0 +1,189 @@
...
+                >
+                  Top of page / Haut de la page
+                  <FontAwesomeIcon
+                    icon={icon['chevron-up']}
+                    className="text-sm sm:hidden pl-2"
+                  />
+                </a>
+              </div>
+            ) : (
+              <section className="md:flex-row md:flex-row flex items-center">
```
shewood, line 70: md:flex-row is in two times?

krischarbonneau: good catch!

shewood (approved):
LGTM!

PR #392, opened 2023-09-28
Removing DS Date.

nibivi77:
Adding our own Date component.


components/Date.js
```diff
@@ -0,0 +1,40 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+export function Date(props) {
+  const { id, label, date } = props
+  const dateFormatted = date ? date.replace(/^(.{4})(.{2})/gm, '$1-$2-') : 'NA'
+  return (
+    <dl id={id} data-testid={id} className="ds-mt-8 ds-py-2 ds-font-body">
```
shewood, line 8: You will need to remove the ds prefix here for the classes

nibivi77: Good catch

components/Date.js
```diff
@@ -0,0 +1,36 @@
+import PropTypes from 'prop-types'
+
+export function Date(props) {
+  const { id, label, date } = props
+  console.log(date)
```
krischarbonneau, line 5: Stray console log :)

krischarbonneau (approved):
Looks good!

PR #395, opened 2023-10-03
Restoring Middleware

lhr-git:

Fixed AB#
Middleware works now

Restoring the MW and removing the rewrites

If the (test branch) server works/launches then the middleware is working as intended.


nibivi77 (approved):
Just need to merge the branch with dev and it should be good!

PR #409, opened 2023-10-24
Reworking server-side errors.

nibivi77:
## [ADO-146355](https://dev.azure.com/VP-BD/DECD/_workitems/edit/146355)

Fixed [AB#146355](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/146355)

### Notes
 - This is only for the Dashboard page. If the team agrees with his method, I will apply it for all other pages.
 - For now, error 500 is hardcoded. Apollo does not provide an error code with the `.catch(error)`.
 
### Question
- Do we agree with this method? I am open for improvements.


krischarbonneau (approved):
LGTM!

PR #421, opened 2023-10-31
Adding Tracing to Middleware

lhr-git:
- Adding tracing into the mw via instrumentation
- removing the automated sdk method via tracing.js
- removing tracing.js
- Added supporting ENV var to example

- you would need to setup a tracing stack of some sort that supported open telemetry (Grafana/tempo is probably the easiest)

### Additional Notes
- for setting up on dynatrace you can see Senor's Journey or our Next-Template, but basically you just need to add in the ENV vars with the OTEL API KEY and the endpoints for metrics and tracing. SJ has a working example of both.


krischarbonneau (approved):
LGTM

PR #423, opened 2023-11-01
Fixing AEM new content.

nibivi77:
Small AEM fixing.

### Notes
 - Still getting 1 icon error on the dashboard as its not set in AEM yet. 
 
image


shewood (approved):
LGTM

PR #429, opened 2023-11-06
Added missing icon.

nibivi77:
Added missing Icon.


shewood (approved):
Icon shows as expected

PR #444, opened 2023-11-15
update links to open in new tab

shewood:
## [ADO-152384(https://dev.azure.com/VP-BD/DECD/_workitems/edit/152384)

Fixes [AB#152384](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/152384)

Change the new tab exceptions links for contact section rows, most requested tasks
Changed the button on the contact us page to a link with an icon to reflect what it actually is. 

links on contact us pages goes to open href in a new tab.

### DIFF
--- a/components/BenefitTasks.js
+++ b/components/BenefitTasks.js
@@ -32,6 +32,9 @@ export default function BenefitTasks(props) {
                 target={
                   newTabTaskExceptions.includes(task.link) ? '_blank' : '_self'
                 }
+                rel={
+                  props.newTabTaskExceptions ? 'noopener noreferrer' : undefined
+                }
                 onClick={(e) => {
                   //check for exit beta popup flag and not a new tab link, else keep default anchor behavior
                   if (
--- a/components/BenefitTasks.js
+++ b/components/BenefitTasks.js
@@ -33,7 +33,9 @@ export default function BenefitTasks(props) {
                   newTabTaskExceptions.includes(task.link) ? '_blank' : '_self'
                 }
                 rel={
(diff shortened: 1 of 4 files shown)


components/BenefitTasks.js
```diff
@@ -32,6 +32,9 @@ export default function BenefitTasks(props) {
                 target={
                   newTabTaskExceptions.includes(task.link) ? '_blank' : '_self'
                 }
+                rel={
+                  props.newTabTaskExceptions ? 'noopener noreferrer' : undefined
```
nibivi77, line 36: I don't think `props.` should be there right? Just `newTabTaskExceptions`.

components/MostReqTasks.js
```diff
@@ -26,6 +26,9 @@ export default function MostReqTasks(props) {
                 target={
                   newTabTaskExceptions.includes(task.link) ? '_blank' : '_self'
                 }
+                rel={
+                  props.newTabTaskExceptions ? 'noopener noreferrer' : undefined
```
nibivi77, line 30: same here `props.`

nibivi77 (changes requested):

components/BenefitTasks.js
shewood: good catch

components/MostReqTasks.js
shewood: good catch!

nibivi77 (approved):

PR #446, opened 2023-11-16
Integrating OTEL with Dynatrace on DSHP

lhr-git:
Add tracing to SCCH via OTEL configs in the helmfile and helm/k8s secrets
Fix a number of URLs to point to DSHP instead of RHP
Account for odd SRE error with namespaces


krischarbonneau (approved):
LGTM!

PR #467, opened 2023-11-27
Adding required defaults to Dockerfile

lhr-git:
Add defaults to Dockerfile, the build fails without them, and should they need to be overriden in pipeline, that's still possible.
Also removed the pem from the Dockerfile, it seems to only be used in local dev, meaning we shouldn't have it in the docker contianer.

Does docker build without requiring build args? if yes. This is successful.


krischarbonneau (approved):
LGTM

PR #472, opened 2023-11-28
Autoscaler, deployment env, logging changes

lhr-git:
- Adding in autoscaler
- Env vars to deployments
- Secrets in k8s
- Loading logging env var at runtime, previously was only instantiated at build


krischarbonneau (approved):
LGTM

PR #476, opened 2023-11-29
Update to use getServerSession and remove some tests

krischarbonneau:
## [ADO-153418](https://dev.azure.com/VP-BD/DECD/_workitems/edit/153418)

Fixed [AB#153418](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/153418)

### Changelog
fix:Update auth functions to use getServerSession instead of getSession

This was originally addressed in #460 but this PR serves to be a cleaner implementation while also fixing the testing issues present in that PR.

From #460:
This PR updates two of our Auth functions (AuthIsValid and getLogoutURL) to take in session as a parameter instead of accessing session information from the getSession function. The getSession function was never meant to be consumed server side even though it does technically work. 

As a result of these changes, our unit tests for our pages that tested `getServerSideProps` broke as presumably `getServerSession` needs to be mocked. This does not seem very straightforward to do. Upon some inspection however, the tests for `getServerSideProps` aren't really providing any value at this time given they are really only testing the metadata as we are mocking every other function and returning an empty object (and testing if said empty object is there). We could test this metadata on the client side so at this time I don't think the `getServerSideProps` test is needed. 

1. Pull in branch
2. Type `npm run build` and `npm run start`
3. Ensure app runs as expected

### Additional Notes
I also fixed a couple component mocks in our tests that were complaining due to some esLint errors


nibivi77 (approved):
This works for me! Thank you again Kris.

PR #478, opened 2023-11-30
Sch splash edit

shewood:

Fixed AB#

This is only for internal use
Update the splash page to include the image in the background to make it look more like a gc site.


nibivi77 (approved):
:)

PR #487, opened 2023-12-01
Updated AA script.

nibivi77:
## [Task](https://dev.azure.com/VP-BD/DECD/_workitems/edit/158366)

Not sure if this is how it should be done.

Let me know if this si not what as requested.


krischarbonneau (approved):
LGTM

PR #504, opened 2023-12-07
updated and added acronyms

shewood:
## [ADO-159511](https://dev.azure.com/VP-BD/DECD/_workitems/edit/159511)

Fixed AB#159511

Added acronyms for CAL and SIN
Isabelle removed extra space in OAS

AA tags should now work.

### DIFF
--- a/lib/acronym.js
+++ b/lib/acronym.js
@@ -21,6 +21,14 @@ export const acronym = (longText) => {
       return 'OASGIS'
     case 'sécurité de la vieillesse':
       return 'OASGIS'
+    case 'social insurance number':
+      return 'SINOM'
+    case 'numéro d’assurance sociale':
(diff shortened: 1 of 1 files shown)


__tests__/components/Layout.test.js
```diff
@@ -94,7 +94,7 @@ describe('Layout with default text', () => {
         locale="en"
         meta={meta}
         display={display}
-        children={aaChildrenProps}
+        // children={aaChildrenProps}
```
krischarbonneau, line 97: This is fine for the time being but we should really look at properly fixing the component instead of commenting this out

shewood: Yes, and this test was not even being run.

krischarbonneau (approved):

PR #509, opened 2023-12-11
Adding the correct contact AA text.

nibivi77:
Adding the correct AA text to contact us pages.


pages/contact-us/index.js
```diff
@@ -78,6 +78,7 @@ export default function ContactLanding(props) {
                 href={`/${props.locale}/${
                   props.content.pageName
                 }/${link.linkDestination.split('/').pop()}`}
+                data-gc-analytics-customclick={`ESDC-EDSC:${link.linkTitle}`}
```
shewood, line 81: the contact us needs to be added in here something like this. 
data-gc-analytics-customclick={`ESDC-EDSC:contact-us:${link.linkTitle}`}

PR #514, opened 2023-12-18
Update refresh msca

krischarbonneau:
## [ADO-159489](https://dev.azure.com/VP-BD/DECD/_workitems/edit/159489)

Fixed [AB#159489](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/159489)

### Changelog 
fix:Update refresh-msca api to handle bad/expired session

This PR updates our refresh-msca handler to account for when the ECAS `validateSession` endpoint returns false (signifying an expired session). I've updated our auth helper functions to return the value and then interpret the value in the handler to determine which response to send back. I've also added a random id generated on each request so that Next doesn't cache the response. This ensures that the request is being made each time as the response will always be different. If the response coming from ECAS is false, a 401 is returned. This is then interpreted on our pages and the user is redirected to the logout page.

1. Pull in branch
2. Type `npm run dev`
3. Ensure the app runs as expected

### Additional Notes
Currently the refresh happens whenever a page is loaded or on click events every 15 seconds. Should this time be increased to 30 seconds or 1 minute for example so as to try and mitigate their endpoint being overloaded?


pages/api/refresh-msca.js
```diff
@@ -3,25 +3,41 @@
...
-      res.status(200).json({ success: true })
+      //If auth session is valid, make GET request to validateSession endpoint
+      const sessionValid = await ValidateSession(
+        process.env.CLIENT_ID,
+        token.sub
+      )
+      if (sessionValid) {
+        res.status(200).json({ success: sessionValid, id: id })
+      } else {
+        res.status(302).json({ success: sessionValid, id: id })
```
lhr-git: Am I reading this correctly: if the session is not valid, return a 302: found?

krischarbonneau: Updated to `401 unauthorized` based on your comments which makes more sense. I had originally used 302 for a temporary redirect

lhr-git (approved):

PR #536, opened 2024-01-11
Converting Spinner and PageLink to TS.

nibivi77:
Converting Spinner.js and PageLink.js to TS.


components/PageLink.tsx
```diff
@@ -0,0 +1,63 @@
...
+  accessText,
+  href,
+  dataCy,
+  buttonHref,
+  buttonId,
+  buttonLinkText,
+  refPageAA,
+  dashId,
+}: PageLinkProps) => {
+  const linkID = linkText ? linkText.replace(/\s+/g, '') : linkText
```
krischarbonneau, line 28: While I agree with removing the extra id props, I think this should still be something along the lines of:
```suggestion
    const linkID = linkText.replace(/\s+/g, '')
```
Which is pretty much the same as it was before but removes the null check (which isn't needed given `linkText` is a required attribute). The ternary here seems a bit unnecessary.

krischarbonneau: I see now that you changed this because the jest test was failing, but the reason the jest test was failing is because the `has no a11y violations` test is trying to render `PageLink` without supplying any of its required props (such as linkText). Changing the render to be:
```
    <PageLink
      lookingForText="title"
      accessText="accessText"
      linkText="Link text"
      buttonLinkText="buttonLinkText"
    />
```
fixes the issue.

krischarbonneau (approved):
LGTM!

PR #544, opened 2024-01-17
Convert Header to TypeScript.

nibivi77:
image

These two conditional statements have been removed because TypeScript interfaces require the props to be there.

Also, `hasNoMenu` has been completely removed as it was never used.


krischarbonneau (approved):
Works well, great job!

PR #547, opened 2024-01-24
Adjust Pod Specs

lhr-git:
Adjusting the pod specs to better match the actual resources in use under load.

- Increased cpu baseline, reduced memory baseline
- Increased cpu limit, reduced memory limit
- Increased number of initial pods in prod, in order to maintain response time while more pods spin up under load
- Increased CPU usage threshold. Previously replication was triggered too easily, resulting in unnecessary pod volume increases.


krischarbonneau (approved):
LGTM!

PR #561, opened 2024-02-06
Update external links to not use modal

krischarbonneau:
## [ADO-171176](https://dev.azure.com/VP-BD/DECD/_workitems/edit/171176)

Fixed [AB#171176](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/171176)

### Changelog - 
fix:update external links so they no longer result in a modal popup

While the intention is to replace the existing modal entirely as we continue to update the project to TypeScript, this PR serves to accomplish the requirement of removing the modal pop up functionality from all of the program links on the dashboard, as well as the external links on the profile, security settings, and decisions review page in the interim. This PR also updates the Button component styling slightly as it was defaulting to link styling if an href was supplied. This component should probably be re-worked given values for some of the styles aren't even present in our project (eg. `focus:ring-bg-deep-blue-focus` on the primary styling)

1. Pull in branch
2. Type `npm run dev`
3. Ensure the program links and the links on the pages mentioned above do not result in a modal opening. Also ensure the that the modal for the banner still works as expected.


shewood (approved):
LGTM

PR #572, opened 2024-02-23
update many dependencies

shewood:
## [ADO-179207](https://dev.azure.com/VP-BD/DECD/_workitems/edit/179207)

[AB#179207](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/179207)

updated many dependencies and devDependencies. 

there was a BREAKING CHANGE in jest-dom
"Removes the extend-expect script. Users should use
the default import path" so test that used that needed to be updated.

Application works as before.


krischarbonneau (approved):
LGTM

PR #608, opened 2024-04-11
update most required card view

shewood:
## [ADO-184080](https://dev.azure.com/VP-BD/DECD/_workitems/edit/184080)

[AB#184080](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/184080)

If there are 3 or fewer most requested items show them in one column
If there are more than 3, show them in 2 columns with 1,2,3 in the first column then overflow into the second column.
There is a provision with the auto-cols-fr to add a 3rd column of equal size should there be more than 6.


krischarbonneau (approved):
LGTM

PR #622, opened 2024-04-22
192137 remove banner

krischarbonneau:
## [ADO-192137](https://dev.azure.com/VP-BD/DECD/_workitems/edit/192137)

Fixed [AB#192137](https://dev.azure.com/VP-BD/d5ca3cd4-19cb-4c8c-b3e6-a7068f4677e4/_workitems/edit/192137)

This PR removes the `PhaseBanner` component as it is no longer being used. This PR also gets rid of the old `MultiModal` component as we are only using a single modal for timeout purposes now and as such has been replaced by the new `IdleTimeout` component. This PR also removes the 3 unused mappers and all references to them as they are no longer needed. I have chosen to use local translations for the timeout modal for now as AEM was giving me headaches but we can always explore switching back to using AEM at a later date for this.

1. Pull in branch
2. Set the values for `timeout` and `promptBeforeIdle` in the `IdleTimeout` component to something really small (setting the last number to something like 15 as an example)
3. Type `npm run dev`
4. Navigate to the dashboard and don't move your mouse until the modal pops up. Ensure it works as expected
5. Ensure that the banner is not showing on any of the pages

### Additional Notes
The Layout unit tests are still being skipped due to issues with how the AA prefix is passed down. This will need to be looked into.


components/IdleTimeout.tsx
```diff
@@ -0,0 +1,139 @@
...
+              id={`${id}-timeout-modal-header`}
+            >
+              {t.bannerHeading}
+            </div>
+            <button
+              data-cy="x-button"
+              type="button"
+              aria-label={'Close Modal'}
+              onClick={handleOnIdleContinueSession}
+              data-gc-analytics-customclick={`ESDC-EDSC:${refPageAA}:Close-Fermer`}
```
shewood, line 89: Can you remove the ESDC-EDSC from here as it gets added later with the refPageAA

components/IdleTimeout.tsx
```diff
@@ -0,0 +1,139 @@
...
+              </p>
+            </div>
+          </div>
+
+          <div className="flex pt-10 md:space-x-12">
+            <Button
+              id="sign-out"
+              text={t.signOutLinkText}
+              onClick={handleOnIdle}
+              style="secondary"
```
shewood, line 121: can you pass the AA ref? Right now it's showing as undefined.

shewood (commented):
Looks good but a couple of changes for AA needed.

components/IdleTimeout.tsx
krischarbonneau: Good call!

components/IdleTimeout.tsx
krischarbonneau: :+1:

shewood (approved):
Nice.

PR #625, opened 2024-04-23
Temporarily update dashboard mapper for Curam test deployment

krischarbonneau:
### Changelog
fix:Temporarily update dashboard mapper to point to correct links

The latest update to the mapper on dev was to specifically show stream 3 MEIIO links if the `environment` env was set to `development`. This will suffice for our testing on stream 3 but not for other streams since the alternate MEIIO links are exclusive to stream 3. This PR temporarily reverts the change to the mapper so that the old links are used (which matches prod and the other streams) until we update our mapper to account for all streams.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to dashboard and ensure the MEIIO links are being shown correctly.


shewood (approved):
LGTM

PR #653, opened 2024-05-07
Local Settings for Cert Setup for Review

lhr-git:
- Removes the hardcoded path from the MSCA NG API cert
- Adds in the option to disable cert loading for auth disable environments

This shouldn't actually disrupt anything other than on local you can now load a cert from another location (say, locally from ./env.crt)


krischarbonneau (approved):
LGTM.

PR #707, opened 2024-07-16
Add Root Cert Structure to Dockerfile

lhr-git:
Adds the Root cert to the container, allowing for a set chain in the container. If you can cat out the root and the user certs when sh'd into the container, and they differ, then it's worked as intended.


krischarbonneau (approved):
LGTM

PR #708, opened 2024-07-26
Update main from staging (July 26th, 2024)

krischarbonneau:
### Changelog
fix:Updated footer ternary to use publicRuntimeConfig value
fix:Update contact us mapper to accommodate SIN link
fix:Update Dockerfile to include root cert

This PR updates the main branch to be in line with the changes on staging as of July 26th, 2024. This includes the fix for the footer links, the change to the contact us mapper so the SIN link shows up, and the change to the Dockerfile to include the root cert so our MSCA-NG call works on prod.

1. Pull in branch
2. Type `npm run dev`
3. Ensure the footer links are redirecting where they should and the contact us page shows a SIN link


Dockerfile
```diff
@@ -26,7 +28,19 @@ ENV NODE_ENV=production
...
+chmod 644 /usr/local/share/ca-certificates/env.crt && \
+mkdir -p  /etc/ssl/certs/ && \
+echo ${HOSTALIAS_ROOT_CERT} | \
+sed 's/\\n/\n/g' | \
+xargs > \
+/etc/ssl/certs/root.crt && \
+chmod 644  /etc/ssl/certs/root.crt && \
+npm run build
 
 FROM node:20-alpine3.20 AS production
```
gregory-j-baker, line 45: I know this isn't my project, but I accidentally clicked on this PR so thought I'd take a peek.

You should really pin the base image version, for build reproducibility.

```
FROM node:20.16.0-alpine3.19 AS production
```

If you don't, the build you get is not guaranteed to be the same each time.

krischarbonneau: Fair point, thanks for bringing it up. Should be updated now.

lhr-git (approved):

PR #730, opened 2024-09-12
Disable oas card temp links

jhslater:
## [276602](https://dev.azure.com/VP-BD/DECD/_workitems/edit/276602)

Fixed AB#

Comment out OAS Curam temporary links 

### What to test for/How to test:
Test to confirm that there is only one link on the OAS on CURAM MSCA-D Card - ACCOUNT


shewood (changes requested):
this shouldn't be going into dev i don't think

shewood (approved):
LGTM

PR #736, opened 2024-10-09
Use next/script and move AA script out of Head

Charles-Pham:
## [ADO-280112](https://dev.azure.com/VP-BD/DECD/_workitems/edit/280112)

Work towards and potential fix for 280112

Changed to not use `<script>` to import AA in Head since this isn't allowed (https://nextjs.org/docs/messages/no-script-tags-in-head-component) and moved it outside and used `next/script` instead.

AA script should appear in the body instead of the head. Also, on monitoring the network activity there should only be one download of the AA script instead of two.

### DIFF
--- a/components/MetaData.tsx
+++ b/components/MetaData.tsx
@@ -1,4 +1,5 @@
 import Head from 'next/head'
+import Script from 'next/script'
 
 interface Content {
   title: string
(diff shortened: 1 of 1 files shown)


jhslater (approved):
Looks good to me!

PR #737, opened 2024-10-10
Reduce replicaCount and minReplicas to 1 for dev environments.

jhslater:
Change `replicaCount` and `minReplica` settings in dev. When deployed, we should also change `HPA_MAX_REPLICAS` env var in teamcity to a suitable number.

This change will reduce costs for the dev environment. Given that we (will) have so many deployed instances of MSCA-D, and additional resources increase the cost of hosting, we should use as few replications are are necessary to support the type of testing therein. For Performance testing, we would need to ensure that the maximum number of replicas is high enough to support the expected load.


Charles-Pham (approved):
Looks good to me

PR #738, opened 2024-10-17
Load AA script before interactive to see if there's a race condition

Charles-Pham:
## [ADO-280112](https://dev.azure.com/VP-BD/DECD/_workitems/edit/280112)

Additional work to attempt to fix double AA count.

Change to load the AA script as soon as possible. Currently, the page waits until it is interactive.

No impact on page, but if it fixes the problem then AA should be able to see the initial page load.

### Additional Notes
Testing of previous version of code indicates that AA doesn't get the page load when logging in or refreshing, but does when navigating from page to page. Given that seemingly everything else works fine and local testing shows no problems, this suggests some sort of race condition when first loading the page.

### DIFF
--- a/components/MetaData.tsx
+++ b/components/MetaData.tsx
@@ -62,12 +62,12 @@ const MetaData = ({ language, data }: MetaDataProps) => {
 
       {process.env.ENVIRONMENT === 'production' ? (
         <Script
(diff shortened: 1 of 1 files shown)


jhslater (approved):
Looks good to me!

PR #739, opened 2024-10-18
Tests for AA script duplication 

Charles-Pham:

Add test that the AA scripts load only once. Couldn't find a way to test the race condition, so added a comment for why AA script needs to be loaded beforeInteractive.

Tests should pass.


jhslater (approved):
LGTM!

PR #742, opened 2024-10-23
Add analytics tags for error pages and don't fire the pageLoad beacon

Charles-Pham:

Add analytics tags for 404 and 500 pages.
Don't fire the pageLoad beacon on those pages.

See analytics fire on the 404/500 pages and the pageLoad doesn't.


jhslater (approved):
LGTM!

PR #743, opened 2024-10-23
push error event

Charles-Pham:

Add an AA push for error pages

See that the AA push occurs


jhslater (approved):
lgtm!

PR #744, opened 2024-10-23
fix aa on language toggle and menus

Charles-Pham:

Add properties for 404 and 500 error pages so that AA information isn't undefined

AA customclick properties on 404 and 500 pages should not have undefined in them anymore.


jhslater (approved):
lgtm

PR #745, opened 2024-10-25
fix aa tags on cdcp

Charles-Pham:

Add acronym decoding for CDCP so that the AA tags show up. Also remove unused import.

The AA-customclick for CDCP should show up as CDCP and not UND


jhslater (approved):
lgtm

PR #747, opened 2024-10-30
accomodate cert chain for raoidc sade environments on new 10 dot star…

jhslater:
The new 10.* IP's (and their DNS) for ECAS RAOIDC come with a new certificate chain that has to be added so that the sade servers can successfully communicate with RAOIDC. These changes add that certificate to the Dockerfile. Note that it is necessary to add a build argument in teamcity for the cert to be included in the container successfully:` --build-arg AUTH_ECAS_CA=%env.AUTH_ECAS_CA%`

There is a corresponding vault entry to house the new certificate (%vault:bdm-dev-secure-client-hub/data/NEXT-AUTH!/AUTH_ECAS_CA%)


Charles-Pham (approved):
looks good

PR #748, opened 2024-10-31
revert AA error event changes

Charles-Pham:

Revert #742 and #743 while retaining test changes since they're more correct.


jhslater (approved):
lgtm

PR #749, opened 2024-11-05
Switch from defaultProps to default parameters. Any string parameter …

Charles-Pham:
…which should not be empty is now defaulted to mscaPlaceholder.

Switch all instances of defaultProps to default parameters. In addition, every instance of a parameter which...

1) Is a string
2) Isn't content put on the page 
3) Isn't solely passed through to another component
4) And is not used in any logic (including its mere existence)

Should now default to mscaPlaceholder. 

defaultProps warnings should not longer appear. mscaPlaceholder should not appear on any page in the HTML.

### DIFF
--- a/components/BackToButton.tsx
+++ b/components/BackToButton.tsx
@@ -12,8 +12,8 @@ const BackToButton = ({
   buttonHref,
   buttonId,
   buttonLinkText,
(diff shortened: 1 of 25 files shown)


jhslater (approved):
lgtm

PR #750, opened 2024-11-08
Fix alerts not appearing

Charles-Pham:

Bugfix for card alerts not appearing. `hasAlert` wasn't being passed in and was always true by the default props (which accidentally didn't get transferred). This is highly misleading since it's dependent on the code itself to not render when there was no alerts, so `hasAlert` has been removed entirely.

Card alerts should now appear


jhslater (approved):
lgtm

PR #751, opened 2024-11-08
accommodate an additional hostalias in helm

jhslater:
This introduces changes to the container to override the dns<->ip relationship to be able to reach ECAS endpoints (RAOIDC, RASCL) with their expected dns but using the new 10.* IP range provided by SSC.

- An additional ssl certificate associated to the 10.* ip's gets added to the container. Note that a new build argument is needed to pull in the new ssl certificate
- An additional ip/dns resolution gets added to /etc/hosts file within the container to override the real dns resolutions on the cluster and ssc network


Charles-Pham (approved):
looks good

PR #753, opened 2024-11-20
Change the CAL link to open in the same tab

abalogh-esdc:
Open the Canada Apprentice Loan application in the same tab

When clicking on the Canada Apprentice Loan application link, the form opens in the same tab instead of a new one.


jhslater (approved):
tested locally, reviewed code changes, looks good to me.

PR #779, opened 2024-12-12
Change profile to typescript

Charles-Pham:

Change profile and related pages to typescript, non-functional change outside of proper breadcrumb handling on error.


susan-esdc (approved):
lgtm

PR #781, opened 2024-12-16
Convert security settings and decision review to typescript

Charles-Pham:

Convert security settings and decision review to typescript. Also make buttons type an optional onClick with the default doing nothing.

No changes should be visible to end users

### DIFF
--- a/pages/decision-reviews.tsx
+++ b/pages/decision-reviews.tsx
@@ -1,8 +1,9 @@
 import PropTypes from 'prop-types'
 import Heading from '../components/Heading'
-import en from '../locales/en'
-import fr from '../locales/fr'
-import { getDecisionReviewsContent } from '../graphql/mappers/decision-reviews'
+import {
+  DecisionReviewContent,
+  getDecisionReviewsContent,
+} from '../graphql/mappers/decision-reviews'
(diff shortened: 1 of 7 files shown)


pages/decision-reviews.tsx
```diff
@@ -156,23 +188,23 @@ export async function getServerSideProps({ req, res, locale }) {
 
   const content = await getDecisionReviewsContent().catch((error) => {
     logger.error(error)
-    return { err: '500' }
+    return { err: '500' } as DecisionReviewContent
   })
```
gregory-j-baker, line 192: The better way to do this is:

``` typescript
const content = await getDecisionReviewsContent().catch((error): DecisionReviewContent => {
  logger.error(error)
  return { err: '500' }
})
```

as it enforces better typechecks.

As implemented in this PR, a strict typecheck isn't being performed, the return value is being *cast* as `DecisionReviewContent` which could lead to future bugs.

Charles-Pham: fixed in 90106802950b140e4bf12afbe8a110835ff55df6

susan-esdc (approved):
looks fine. I guess lint did something again with commas, but nothing destructive

PR #783, opened 2024-12-19
Allow AEM folder to be specified

Charles-Pham:

Support using the AEM preview folder

No visible functional changes

### Additional Notes

Requires an additional variable to be set in the config


jhslater (approved):
lgtm

PR #784, opened 2024-12-31
Convert layout to typescript

Charles-Pham:

Convert layout to typescript

No functional changes should be visible to end user (some vestigal code removed)

### DIFF
--- a/components/Layout.tsx
+++ b/components/Layout.tsx
@@ -2,49 +2,40 @@ import throttle from 'lodash.throttle'
 import { signOut } from 'next-auth/react'
 import getConfig from 'next/config'
 import { useRouter } from 'next/router'
-import PropTypes from 'prop-types'
-import { useCallback, useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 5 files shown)


components/Layout.tsx
```diff
@@ -2,49 +2,40 @@ import throttle from 'lodash.throttle'
...
-import MetaData from './MetaData'
+import MetaData, { Data } from './MetaData'
 
-export default function Layout(
-  props = {
-    locale: 'en',
-    meta: '',
-    langToggleLink: '',
-    breadCrumbItems: [],
-    bannerContent: {
```
susan-esdc, line 21: Any reason why bannerContent is removed? (i may be missing something)

susan-esdc: I guess it's not being used

susan-esdc (approved):
lgtm

components/Layout.tsx
Charles-Pham: Yes, this was to support the banner while it was still in beta, so it's no longer actually used anywhere.

PR #787, opened 2025-01-08
Convert app, refresh, and metrics to typescript

Charles-Pham:

Convert _app, refresh-msca, and metrics to typescript.

No functional changes should be visible to end user.


abalogh-esdc (approved):
Looks good to me.

PR #797, opened 2025-01-24
AA fixes

Charles-Pham:

Changes the AA event pushed on error pages to an error instead of a pageload.
Remove the page name from AA to harmonize french and english tags
Change AA tag for Most requested to harmonize french and english 

AA beacons all fire properly


jhslater (approved):
lgtm

PR #830, opened 2025-03-12
Target CPU and Memory utilizations adjusted to 70

jhslater:
This change sets a lower HPA CPU and Memory threshold for the pods so that horizontal pod autoscaling happens earlier, before performance has degraded to the point where users could get errors and timeouts.


Charles-Pham (approved):
looks fine

PR #833, opened 2025-03-17
Switch from next auth to panva node oidc 

susan-esdc:

Fixed AB#

1. Create an openIdClientService with 2 methods - authorize and callback. The openIdClient within this service includes a jwks set which contains 1 private/public key pair. The private key is contained in vault while the public one is in /public/jwk/public.jwk
2. The call to authorize from the login page reaches ECAS' /authorize endpoint and sends it various checks such as nonce, state, and redirect url. If the user isn't logged in on ECAS, they are sent to the login page and redirected back to msca dashboard once they are done. Otherwise, they are redirected back to the AUTH_REDIRECT_ENDPOINT (/api/oauth-callback) with a code in the url parameters.
3. Within /api/oauthcallback, the call to the open id service's callback method is made - sending it the code it retrieved from the url parameters (req.query) as well as a client assertion token. This client assertion token is signed with our private key and validated by ECAS with our public key. In exchange, ECAS sends back a tokenSet contained an access_token and an id_token. The access_token is encrypted and not decrypted by us. It's used to call ECAS' userinfo method (called with axios). The idToken was encrypted by ECAS with our public key and decrypted by us with our private key. The idToken contains the session Id and is added to a cookie.
4. Whenever a user clicks a new page or refreshes a page or there is a click or visibility change event (as coded in Layout.tsx), a call to ValidateSession in auth.ts is made. This method makes an axios call to the /validatesession endpoint in ECAS, sending it the sessionId retrieved from the sessionId cookie. If the session is valid, the cookie expiry time is extended, otherwise the user is redirected to login


pages/api/oauth-callback.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  updateMscaNg(
+    decryptedUserInfoToken.sin as string,
+    decryptedUserInfoToken.uid as string,
+  )
+  const locale = getCookieValue('localeForOauthCallback', req.cookies)
+  deleteCookieWithName(req, res, 'localeForOauthCallback')
+  const dashboardRedirect =
+    locale === 'en'
+      ? 'https://mscad-sys2-s2.bdm.dshp-phdn.net/en/my-dashboard'
```
jhslater, line 91: Could these url's be picked up from config instead of hard-coding?

pages/api/openid-client-service.ts
```diff
@@ -0,0 +1,150 @@
...
+      state: string,
+      nonce: string,
+      codeVerifier: string,
+      jti: string,
+      expiry: number,
+      iat: number,
+      nbf: number,
+    ) => {
+      return openIdClient.callback(
+        'https://mscad-sys2-s2.bdm.dshp-phdn.net/api/oauth-callback',
```
jhslater, line 64: domain will vary by deployment instance, should be configurable

pages/api/openid-client-service.ts
```diff
@@ -0,0 +1,150 @@
...
+
+const log = getLogger('openid-client-service')
+
+/**
+ * Return a singleton instance (by means of memomization) of the openid client service.
+ */
+export const getOpenIdClientService = moize.promise(createOpenIdClientService, {
+  onCacheAdd: () => log.info('Creating new open id client service'),
+})
+console.log('testing')
```
Charles-Pham, line 14: these should be removed

pages/api/openid-client-service.ts
```diff
@@ -0,0 +1,150 @@
...
+console.log(process.env.AUTH_PRIVATE)
+const authPrivate = process.env.AUTH_PRIVATE
+
+let jwk = {} as OpenIdJoseJWK
```
Charles-Pham, line 18: is this cast necessary? It seems like it should fit the interface

pages/api/openid-client-service.ts
susan-esdc: fixed - used path instead of full url

lib/auth.ts
```diff
@@ -50,38 +44,55 @@ export async function ValidateSession(clientId, sharedSessionId) {
             },
           )
           .then((response) => response)
-          .catch((error) => logger.error(error))
+          .catch((error) => {
+            if (error.response) {
+              console.error(error.response.data)
```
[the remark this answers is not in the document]
susan-esdc: fixed - used logger

lib/auth.ts
```diff
@@ -50,38 +44,55 @@ export async function ValidateSession(clientId, sharedSessionId) {
...
-  if (session && token) {
+  const sessionId = getCookieValue(
+    process.env.AUTH_COOKIE_PREFIX + 'sessionId',
+    cookies,
+  )
+  if (sessionId !== undefined && sessionId !== null && sessionId !== '') {
     return (
       process.env.AUTH_ECAS_GLOBAL_LOGOUT_URL +
-      `?client_id=${process.env.CLIENT_ID}&shared_session_id=${token.sub}&ui_locales=${localeParam}`
+      `?client_id=${process.env.CLIENT_ID}&shared_session_id=${sessionId as string}&ui_locales=${localeParam}`
```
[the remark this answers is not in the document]
susan-esdc: fixed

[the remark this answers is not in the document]
susan-esdc: fixed

pages/api/openid-client-service.ts
susan-esdc: fixed

pages/api/openid-client-service.ts
susan-esdc: fixed

jhslater (approved):
LGTM

PR #850, opened 2025-04-04
change usa back to canada!

susan-esdc:

Fixed AB#


jhslater (approved):
We do not want to be annexed, thanks!

PR #854, opened 2025-04-08
[Snyk] Upgrade @opentelemetry/sdk-node from 0.57.2 to 0.200.0

krischarbonneau:
snyk-top-banner

<h3>Snyk has created this PR to upgrade @opentelemetry/sdk-node from 0.57.2 to 0.200.0.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.

<hr/>

- The recommended version is **4 versions** ahead of your current version.

- The recommended version was released **22 days ago**.

<details>
<summary><b>Release notes</b></summary>
<br/>
  <details>
    <summary>Package name: <b>@opentelemetry/sdk-node</b></summary>
    <ul>
      <li>
        <b>0.200.0</b> - <a href="https://redirect.github.com/open-telemetry/opentelemetry-js/releases/tag/experimental%2Fv0.200.0">2025-03-17</a></br><h2>0.200.0</h2>
<h3>Summary</h3>
<ul>
<li>The <strong>minimum supported Node.js has been raised to <code>^18.19.0 || &gt;=20.6.0</code></strong>. This means that support for Node.js 14 and 16 has been dropped.</li>
<li>The <strong>minimum supported TypeScript version has been raised to 5.0.4</strong>.</li>
<li>The <strong>compilation target for transpiled TypeScript has been raised to ES2022</strong> (from ES2017).</li>
<li>The <strong>public interface has changed</strong>
<ul>
<li>for notes on migrating to 2.x / 0.200.x see <a href="https://redirect.github.com/open-telemetry/opentelemetry-js/tree/main/doc/upgrade-to-2.x.md">the upgrade guide</a></li>
</ul>
</li>
<li>Only stable versions <code>2.0.0</code> are compatible with this release</li>
</ul>
<h3>💥 Breaking Change</h3>
<ul>
<li>feat(exporter-prometheus)!: stop the using <code>type</code> field to enforce naming conventions <a href="https://redirect.github.com/open-telemetry/opentelemetry-js/pull/5291" data-hovercard-type="pull_request" data-hovercard-url="/open-telemetry/opentelemetry-js/pull/5291/hovercard">#5291</a> <a class="user-mention notranslate" data-hovercard-type="user" data-hovercard-url="/users/chancancode/hovercard" data-octo-click="hovercard-link-click" data-octo-dimensions="link_type:self" href="https://redirect.github.com/chancancode">@ chancancode</a>
[...]

PR #861, opened 2025-05-16
update open telemetry

Charles-Pham:
#853 and #854, plus an additional package to accommodate those two, as well as updates to the resource API usage. Also update to less deprecated attributes.


MeanITTy:
Hey there!

Sounds like you're diving into some important updates with OpenTelemetry—#853 and #854 definitely seem crucial. It's great that you're also tackling the additional package and those API updates. Keeping up with less deprecated attributes is always a good move.

If you ever need any extra help or a fresh pair of eyes on this, Mean IT Consulting LLC has some solid folks who can lend a hand. Just PM if you're interested, and we can chat more about it.

Cheers!

